PR #8, opened 2020-05-16
add markdown support

### LINKED ISSUE #2: Add markdown support
Helpfiles are currently writte in MD. This makes them look good in github but not so much when viewed in the client.

Using this lib: https://github.com/netvl/md.rs we could perhaps improve the look of the markdown files when shown in the client.

Coloring is out, but the other data should turn out better then what we have at least.

tarkah:
resolves #2 

I gave `mdcat` a shot and it looks great out of the box! I'm sure you can mess around with the settings a bit, as you'd know best what terminal capabilities should be enabled and what parsing options to enable.


LiquidityC:
Looks really nice at first glance. I’ll check it out once I get to my laptop. Thanks

src/help_handler.rs
```diff
@@ -13,15 +19,34 @@ impl HelpHandler {
...
+            }
         } else {
-            self.writer
-                .send(Event::Info("No such helpfile found".to_string()))
-                .unwrap();
-        }
+            "No such helpfile found".to_string()
+        };
+
+        self.writer.send(Event::Info(output)).unwrap();
```
LiquidityC, line 49: Use Event::Output instead. Event::Info prefixes with [**] which doesn't really help here.

tarkah:
Good call, updated.

When text is sent to the terminal that is longer than the height of the viewport, it currently puts you at the end of the text. So when you use `/help scripting` you need to scroll back up to get to the top of the help page. It'd be nice if it placed the top of the page at the top of the viewport, and then required you to page down / end to the end? Maybe there can also be a visual cue at the bottom of the viewport (an ellipsis?) that indicates there is more text to scroll though.

LiquidityC:
Looks good. As for your suggestions regarding scrolling and indicators. Those are great and things I have in mind. That's why there is a top-bar and a separator. Figure I can use those for what you are asking. Just push the issues on the stack and I'll pick them up as I go along. Or you can pick them up yourself :smile:

PR #20, opened 2020-05-19
add saved servers

### LINKED ISSUE #14: Save connection details
It'd be nice to be able to save the connection details to a server as a friendly name, that can then be used with `/connect` to easily reconnect to that server. And the ability to list all saved connections. Or maybe an alternative would be to have a "Recently Connected" list?

Would be interested in your preference.

tarkah:
resolves #14


tarkah:
`SavedData` trait can be impl'd on any struct that is `Serialize + Deserialize + Default`. You just need to define the relative path (relative to `data_dir/blightmud`) that it will save the serialized struct to.

I've added `anyhow` for easier propagation of errors.

`/connect` will try any saved server names if only 1 argument is passed.

src/event.rs
```diff
@@ -143,6 +150,72 @@ impl EventHandler {
                 screen.redraw_top_bar("", 0)?;
                 Ok(())
             }
+            Event::AddServer(name, connection) => {
```
LiquidityC, line 153: To prevent this method from getting longer. Perhaps create a new one eg. "handle_store_events" which handles these events?

src/save.rs
```diff
@@ -0,0 +1,43 @@
+use anyhow::{format_err, Result};
```
LiquidityC, line 1: The root directory is getting a bit crowded. Perhaps move this to a subfolder? (named: 'io' or 'disk_io' ?).

What do you think? I'm to new to rust to argue for what's common project structuring.

Adding "logging" is in the list of issues which also pertains to disk storage so those files will probably share this folder.

src/connection.rs
```diff
@@ -0,0 +1,27 @@
+use crate::save::SaveData;
```
LiquidityC, line 1: Root directory is getting sort of big. Perhaps move this to a subfolder named 'model' since this is a straight up data object. What do you think? Maybe overkill since we might not get many more data objects to handle? I think we'll eventually have to make one for line etc.

src/save.rs
```diff
@@ -0,0 +1,43 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+
+use std::fs;
+use std::path::PathBuf;
+
+pub trait SaveData: DeserializeOwned + Serialize + Default {
+    fn relative_path() -> PathBuf;
+
+    fn path() -> Result<PathBuf> {
+        let mut data_dir =
```
LiquidityC, line 11: Could we put a conditional here: #[cfg(debug_assertions)]
Eg. when debug (dev dir) just write to the project root. Otherwise write to data dir. This way developing and actually using the software won't interfere? Check how it's done in main.rs for the log dir ;)

LiquidityC: An idea might be to make the logging initiation and this fetch their root dir from the same method so they share the logic. That's probably the best way to do it. Then we can re-use it for future file io.

Cargo.toml
```diff
@@ -22,3 +22,6 @@ flate2 = "1.0.14"
 mdcat = { version = "0.16.1", default-features = false }
 pulldown-cmark = "0.7.1"
 syntect = "4.1.1"
+serde = { version = "1.0", features = ['derive'] }
+ron = "0.5"
+anyhow = "1.0.31"
```
LiquidityC, line 27: Super cool. Never heard of this before. I'll replace all my boxing stuff once this is merged

tarkah: Yeah, `anyhow` is the bomb! If you want to derive custom error types very easily, you should also check out `thiserror`. Between those two, you should be set on error handling.

src/save.rs
tarkah: Yup, I like this idea. We can use `lazy_static` to define a global variable at runtime (needed for `dirs`) and mix in the conditional for debug assertions to use the crate root.

src/event.rs
tarkah: Agreed

src/save.rs
tarkah: Yeah root is definitely getting crowded. I assume you're more of a fan of the folder/ + mod.rs in folder vs folder.rs + folder/ layout then? Otherwise root is just as crowded :P

LiquidityC: I’ve only used the mod.rs variant so far. I find it’s quite a clean structure

tarkah:
Alright, check out all the changes. I added `DATA_DIR` static ref which we can use throughout the program, and is conditional on debug assertions. I also updated `start_logging` to reference this and collapsed into one function w/ the `log_level` being conditional.

PR #39, opened 2020-05-26
Break scripting.md into multiple files.

### LINKED ISSUE #37: Split scripting.md into multiple helpfiles.
The helpfile `resources/help/scripting.md` is really long and bulky. It would be nice if we split this up into several smaller helpfiles and had the main scripting.md point the user to other helpfiles instead.

It's a pretty straight forward task and it can easily be done in increments.

This is the workflow:
- Check scripting.md for something to break out.
- Create a new file with an appropriate name.
- Tell the user about the new helpfile in scripting.md
- Add the new file to `src/ui/helper_handler.rs` so that it can be viewed

Eg. `lua_timers.md`, `lua_triggers.md` etc.

lisdude:
Closes #37

### DIFF
--- a/resources/help/lua_aliases.md
+++ b/resources/help/lua_aliases.md
@@ -1,0 +1,18 @@
+# Aliases
+
+Aliases allow you to trigger a callback function when a certain command is typed.
+
+## Creating an Alias
+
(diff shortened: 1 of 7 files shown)


LiquidityC:
Nice!

PR #68, opened 2020-06-26
Update lua_triggers.md

robg204:
Update existing example trigger to actually work.  Was calling match[2] instead of matches[2] as defined in the callback.

Add an additional example using lua long brackets which do not require tedious extra escaping for Lua strings, useful for pcre definitions.


LiquidityC:
I learned something new today. Awesome, thanks!

robg204:
> I learned something new today. Awesome, thanks!

I've written about 1200 lines of lua so far porting to blight from tintin++, this tidbit was one of the better pieces I figured out!

Issue #76, opened 2020-07-11
Printing to statusline causes... problems


Xandaros:
![](http://tcp.st/kokv9.png)

I'm printing to the status line in response to a GMCP message (Char.Vitals). I reset the colour at the end of the string, just by virtue of how it is generated. The bar does not change its colour back to what it should be after the text - I would expect it to.

After printing to the line, the end of the bar is missing.
The cursor appears to be at the end of the visible bar, not on the input line. Typing anything moves it back down, though.

Complete relevant script:
```lua
local function formatStat(value, max)
    local color
    value = tonumber(value)
    max = tonumber(max)
    if value < 0.3 * max then
        color = C_BRED
    elseif value < 0.7 * max then
        color = C_BYELLOW
    else
        color = C_GREEN
    end
    return color .. value .. "/" .. max .. C_RESET
end

local function updateStatus()
    local vitals = GMCP.char.vitals
    local hp = formatStat(vitals.hp, vitals.maxhp)
    local mp = formatStat(vitals.mp, vitals.maxmp)
    local ep = formatStat(vitals.ep, vitals.maxep)
    local wp = formatStat(vitals.wp, vitals.maxwp)
    local line = hp .. " " .. mp .. " " .. ep .. " " .. wp

    blight:status_line(0, line)
end

blight:on_gmcp_ready(function()
    blight:output("GMCP enabled")
    blight:register_gmcp("Char")
    blight:add_gmcp_receiver("Char.Vitals", function(data)
        obj = json.decode(data)
        GMCP = GMCP or {}
        GMCP.char = GMCP.char or {}
        GMCP.char.vitals = obj
        updateStatus()
    end)
end)
```

LiquidityC:
Yes that’s because you reset the color and it doesn’t handle that. It definitely should handle that so this is a definite bug. I’ll get on it during the week.

LiquidityC:
I just realized you had a PR here. I’ll check that out as soon as I’m back from travels. 😉

Xandaros:
Yeah, this was annoying me enough that Blightmud was essentially unusable for me. The colour is one thing, what really annoyed me was the cursor being in the wrong place. And I had that happen even without any colour codes in the status line, the cursor would end up on the last character of the bottom-most status line.

I figured this should be easy enough to fix myself, so I did. Turned out to be more complicated than expected, but that's how software works sometimes, isn't it :D

PR #77, opened 2020-07-12
Fix status line when used with colour codes

Xandaros:
See #76 
I tried many different ways of fixing this, but in the end, simply pretending the string should be longer than it actually is was the only one that worked.

In particular, I originally wanted to use `String::repeat` to generate the String of line characters that come after the status message. Somehow, the last couple characters were always gone. Even if I made it longer than necessary, the characters would appear in the input area, but at the end of the line there were still characters missing - as if it decided to linebreak early.
No idea what was going on there, so this solution it is.

Tested with a status height of 1 and 2.

### DIFF
--- a/src/ui/screen.rs
+++ b/src/ui/screen.rs
@@ -133,12 +133,25 @@ impl StatusArea {
         )?;
 
         let custom_info = if !custom_info.trim().is_empty() {
-            format!("━ {} ", custom_info.trim())
+            format!(
+                "━ {}{}{} ",
+                custom_info.trim(),
+                color::Fg(color::Reset),
+                color::Fg(color::Green)
+            )
         } else {
             "".to_string()
         };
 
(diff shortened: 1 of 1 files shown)


Xandaros:
Fine, I'll run `cargo fmt`. I think it makes it less readable, though...

src/ui/screen.rs
```diff
@@ -133,12 +133,20 @@ impl StatusArea {
         )?;
 
         let custom_info = if !custom_info.trim().is_empty() {
-            format!("━ {} ", custom_info.trim())
+            format!("━ {}{} ", custom_info.trim(), color::Fg(color::Green))
```
LiquidityC, line 136: Do a color::Fg(color::Reset) here before switching to Green. Just to make sure all the "flags" are cleaned first.

Issue #74, opened 2020-07-11
Crash if Lua script errors in on_connect hook


Xandaros:
In #73 : "Crashing on lua script errors isn't great. Didn't think that happened. If you can recall the errors then please post them."
So here it is.
```lua
blight:on_connect(function(host, port)
    nonexistent()
end)
```
Causes blightmud to crash on connect. Indexing nil values has the same effect, I haven't tried other types of errors.
A report file is created in /tmp:
```
name = 'blightmud'
operating_system = 'unix:Arch'
crate_version = '0.5.0'
explanation = '''
Panic occurred in file 'src/lua/lua_script.rs' at line 219
'''
cause = '''called `Result::unwrap()` on an `Err` value: RuntimeError("[string \"/home/xandaros/.config/blightmud/achaea.lua\"]:39: attempt to call a nil value (global \'nonexistent\')\nstack traceback:\n\t[C]: in global \'nonexistent\'\n\t[string \"/home/xandaros/.config/blightmud/achaea.lua\"]:39: in function \'__connection_callback\'")'''
method = 'Panic'
backtrace = '''

   0: 0x5577aad756aa - <unresolved>
   1: 0x5577aad3641b - <unresolved>
   2: 0x5577aad3492d - <unresolved>
   3: 0x5577aad7ffa3 - <unresolved>
   4: 0x5577ab039108 - <unresolved>
   5: 0x5577aad3a312 - <unresolved>
   6: 0x7fd0e08c4002 - __libc_start_main
   7: 0x5577aad0726e - <unresolved>
   8:        0x0 - <unresolved>'''
```

LiquidityC:
Good spot. I’ll take a look during the summer.

PR #83, opened 2020-07-14
Prevent lua crash on errors in on_connect hook

LiquidityC:
Fixes: #74


Xandaros:
I briefly looked at this and `on_disconnect` might have the same issue. I didn't test it, though

LiquidityC:
Exactly the same problem essentially. Fix is up :smile:

LiquidityC:
Exactly the same problem. Patching and including in this PR

Issue #84, opened 2020-07-14
Screen messes up on resize


Xandaros:
While working on #80, I was looking at the current code and was wondering how exactly the screen is currently drawn.
I noticed that line breaks might not work properly after a resize and I tested it - turns out a lot more messes up than just the line breaks:
![](http://tcp.st/5c61_.png)

Assuming I get #80 done in a reasonable amount of time, I don't think fixing this is very important, since it will be fixed with #80 anyway.

LiquidityC:
The screen is setup using a scroll area. The output is a scroll area and the input prompt is a regular raw terminal. Resizing used to work decently. Not sure how it behaves with history since that was added after. I'll have a look when I get some time.

LiquidityC:
There's a redraw that should be triggered on resizes. I think this needs to be extended to redraw all the bars and the history to completely clean the screen. Shouldn't be to hard to work out.

robg204:
I notice this if I re-attach to my tmux session from another computer.  The status_line updates remain in the middle of my screen, etc, while the actual input and scrolling regions are updated correctly, any updates to status line continue to be placed where they were at client start up.

LiquidityC:
Exactly. I think that's the problem. Before printing to the status area was a thing resizing worked nicely. I most likely broke it when I was working. Resizing isn't a common thing. Hence it's flown under the radar.

LiquidityC:
Found the issue here. Scrolling will still not adapt but I don't consider that a major problem. If you shrink the view you get basic wrapping. If you enlarge you just keep the current wrapping until more output is received from the mud (new output has new width and correct wrapping).

PR #86, opened 2020-07-15
Fixes screen resizing with status area.

LiquidityC:
Fixes: #84


src/main.rs
```diff
@@ -391,10 +391,11 @@ fn run(
                 }
                 Event::Redraw => {
                     screen.setup()?;
-                    screen.reset_scroll()?;
```
LiquidityC, line 394: screen.setup() runs screen.reset_scroll() so this was pointless.

PR #87, opened 2020-07-16
Adds a basic "/lua" command through @robg204's gist

LiquidityC:


src/lua/lua_script.rs
```diff
@@ -38,13 +38,17 @@ fn create_default_lua_state(writer: Sender<Event>, dimensions: (u16, u16)) -> Lu
...
-            ctx.load(bindings).exec()?;
-            let defaults = include_str!("../../resources/lua/defaults.lua");
-            ctx.load(defaults).exec()?;
+
+            ctx.load(include_str!("../../resources/lua/bindings.lua"))
+                .exec()?;
+            ctx.load(include_str!("../../resources/lua/defaults.lua"))
+                .exec()?;
+            ctx.load(include_str!("../../resources/lua/lua_command.lua"))
+                .exec()?;
```
Xandaros, line 51: Should probably make this more general and just load all `resources/lua/*.lua`

LiquidityC: Yup, but that doesn't work with `include_str!()` since that's done at compile. Unless you know something I don't :smile:

Xandaros: Yeah, it's not trivial. You'll probably have to make a custom macro for it.
If you prefer, I can do that, too. Something like `lua_load_directory!(ctx: _, path: &str) -> Result<()>` (Don't know the type of context)

LiquidityC: You are most welcome to give it a shot. This will have to do for the time being.

Issue #80, opened 2020-07-12
Split window


Xandaros:
Something I have always wanted in a MUD client is the ability to have multiple windows up at once in a vim-like fashion.

That way you could have, for example, one main window, a window displaying a map, a window showing channel chatter, etc.

I suggest adding a function for creating windows. Perhaps `blight:add_window()`.

I'm not entirely sure how you would actually output to a different window, though. `blight:output` could take an index as first parameter, but it breaks existing scripts and would be less convenient to use for single-window users.
Maybe a second output function, something like `blight:output_window(number, string...)`?

Finally, there should be an option to resize windows. Capturing mouse events would probably be the most convenient for the user, but that's not entirely necessary for now. Just having functions that resize should be fine.

The main question remaining is how to lay out the windows. I would like a master+stack layout so you have one main window on the left and a column of windows on the right. How the windows are laid out is a choice that should be given to the user, though, and I'm not entirely sure what the best way to do that would be.
One option:
Have a concept of a window group.
There is one main window group, which will be laid out over the entire screen.
A window group can contain windows and other window groups.
A window group can either be horizontal or vertical.

That would essentially be how the window manager i3 (and probably others) does its layout. Another option is using predefined layouts, like, for example, Xmonad handles it.

I'm leaning towards manual. Should be easier to implement and unlike a window manager, I expect the number of windows to remain fairly static throughout a session.

Thoughts? I hope this doesn't go against some design philosophy :smile:

Xandaros:
Possible, concrete API:

`blight:create_window() -> userdata<Window>` - Creates a new Window, but does not display it or anything
`blight:get_main_window() -> userdata<Window>` - Returns the main window
`blight:set_layout(table)` - Sets the layout. The table contains windows or tables at numerical indices. There is also a key `vertical`, which, when truthy, makes the layout vertical instead of horizontal. Subtables follow the same format. If the main window is not part of the layout, it is assumed to be in the first position.

`Window:clear()` - Clears the contents of a window
`Window:output(string...)` - Like `blight:output`, but outputs to the window
`Window:set_size(integer)` - Sets the size of the window in characters. If the size is larger than fits on the screen or the size would cause other windows to go off-screen, the size is ignored. Whether the size is width or height is determined by whether the layout is horizontal or vertical.
`Window:reset_size()` - Resets size. Same as `Window:set_size(nil)`. Windows without size are evenly sized.
[...]

PR #97, opened 2020-07-19
Port to tui-rs

Xandaros:
As discussed on Discord, this makes Blightmud use the tui crate instead of using termion directly.
Added dependencies: tui, ansi-parser

Currently, the widgets are re-created on every render. This is a noticeable slowdown in debug builds, but in release builds this is not noticeable whatsoever.
I'm held off on fixing that, as I want to implement the windowing system as discussed in #80 next. The entire rendering is going to change, so I didn't see any point in fixing that at this point. (Again, not noticeable in release builds)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -328,7 +328,7 @@ fn run(
                 }
                 Event::ScrollUp => screen.scroll_up()?,
                 Event::ScrollDown => screen.scroll_down()?,
-                Event::ScrollBottom => screen.reset_scroll()?,
+                Event::ScrollBottom => screen.reset_scroll(),
                 Event::StatusAreaHeight(height) => screen.set_status_area_height(height)?,
                 Event::StatusLine(index, info) => screen.set_status_line(index, info)?,
                 Event::LoadScript(path) => {
@@ -381,12 +381,13 @@ fn run(
                     session.lua_script.lock().unwrap().remove_timed_function(id);
                 }
                 Event::Redraw => {
-                    screen.setup()?;
+                    let _ = screen.update_size();
                     if let Ok(mut script) = session.lua_script.lock() {
                         script.set_dimensions((screen.width, screen.height))?;
                     }
                     let prompt_input = session.prompt_input.lock().unwrap();
-                    screen.print_prompt_input(&prompt_input, prompt_input.len());
+                    screen.set_cursor_pos(prompt_input.len())?;
+                    screen.set_prompt(&*prompt_input);
                 }
                 Event::Quit => {
                     session.terminate.store(true, Ordering::Relaxed);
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -381,23 +381,23 @@ fn run(
...
                     session.terminate.store(true, Ordering::Relaxed);
                     session.disconnect();
                     break;
                 }
             };
-            screen.flush();
         }
     }
-    screen.reset()?;
+    //screen.reset()?;
```
LiquidityC, line 400: Remove this.

src/ui/screen.rs
```diff
@@ -34,39 +43,94 @@ impl error::Error for TerminalSizeError {
...
+        if self.height() > 1 {
+            // Draw non-bar lines
+            for line_no in 1..(self.height() - 1) {
+                if remaining_height == 0 {
+                    break;
+                }
+                remaining_height -= 1;
+
+                if let Some(line) = &self.status_lines[line_no as usize] {
+                    buf.set_string(area.x, area.y + line_no, line, Style::default());
```
LiquidityC, line 106: Colors aren't handled here so you get odd printout.

src/ui/screen.rs
```diff
@@ -188,182 +162,159 @@ impl History {
...
-        self.status_area.set_height(height, self.height);
-        self.setup()?;
-        Ok(())
-    }
+    pub fn redraw(&mut self) -> Result<()> {
+        let main_window = {
+            let block = Block::default()
+                .borders(Borders::TOP)
+                .border_style(Style::default().fg(Color::Green))
+                .title(self.title.as_str());
```
LiquidityC, line 234: Is it possible to indent the title as it was before? (prepended by one 'bar-char')

LiquidityC:
Apart from the tests. Would it be better if we pushed all this to a new `tui-port` branch. We can then create issues and help out bringing everything up to speed with what we currently have on the stable (master) branch.
I also have a gnawing issue regarding all the extra parsing we need to do. Eg. Parsing every line for color before rendering to screen. I think this is a major downside because it costs a lot of cycles.

Xandaros:
The colour parsing should only happen when the line gets added. The history stores `OwnedSpans` now, which contain the colour information.

src/ui/screen.rs
Xandaros: I changed the border type to double, as it was before. I also prepended TWO bar-chars.
For some reason, if I only prepend one, it ignores it... I looked into the tui code and couldn't see a reason as to why that should be, but that's what it did.
Putting two bar-chars renders both... Don't really get it, but I also don't think it matters overly much.

Issue #91, opened 2020-07-16
Ability to send a line of text and have it processed as if it came from either the mud or the user


robg204:
For easier testing of scripts without needing to wait for the input or output.  It would be nice if we could test this both on mud output (triggers), as well as on user input (aliases).  Extra points if we could also do this from a lua binding, as you could do fun things like write a unit test for your triggers and aliases, which would be amazing :)

LiquidityC:
Request for input here.

I added the lua function `mud_output(output)` which sends a line to the client as if it came from a mud server.
I've then created a macro based on this function called `/test <test line>`

I think the naming of the macro `/test` sounds and looks fine. Question is, do we want to call the lua method `test(line)` as well. Is that confusing or does it make sense? Discuss!!

robg204:
I don't think it should be test(line) in lua, mud_output is okay, there might be a slightly better name, but I'm not really sure.  

Ideally, we would have a similar mechanic for the user input, to allow for things like creating a script to allow command stacking, where stacked commands might need to trip other triggers (unless the stacking element is deemed to be better handled in rust rather than in lua).

LiquidityC:
I'll leave it as `mud_output(str)` with the macro `/test <line>`. I'll continue with the alias part. I'll add the function `user_input(str)` to lua.

PR #103, opened 2020-08-12
Adds the basics for #91

LiquidityC:
Fixes: #91


LiquidityC:
Not complete yet. Need handling for sending user input from lua as well

Issue #90, opened 2020-07-16
Ability to poll trigger list, alias list, timer list


robg204:
Discussed on discord:

Currently, we track aliases, timers, and triggers (I do this in tables), so that we may remove triggers when reloading in place, or for other reasons.  It would be nice to be able to poll the client for a list of active IDs from lua rather than having to track it ourselves.  Useful for situations where we think our environment is wrong.

LiquidityC:
This information is actually hidden in the lua state already. Triggers and aliases are stored in tables. So it should be pretty easy to implement on the lua side. Check `lua_script.rs` for the names of those tables.

robg204:
Today I learned.. :)  Great! Thanks!

LiquidityC:
I think we could still write some more user friendly wrappers for it.

LiquidityC:
I've decided I want to extend the new functionality slightly by returning a table containing `id` as key and a string representation of the `regex` as value where applicable. That should make the information slightly more useful when debugging and what not.

I'll also create some macros based on this such as `/aliases` and `/triggers`

LiquidityC:
This is done now, I extended the following:
- `blight:get_aliases()` and `blight:get_triggers()` returns proper ids and objects in a table
- Added macros `/aliases` and `/triggers`
- Added methods `blight:enable_alias(int, bool)` and `blight:enable_trigger(int, bool)`
- Relevant documentation and tests

PR #104, opened 2020-08-17
Adds methods to get trigger,alias,timer ids

LiquidityC:


LiquidityC:
When working on #90 i realised that it was possible for the user to delete aliases such as `/lua` command etc. I've introduced a method in lua `blight:core_mode(bool)` this should be set to `true` at the start of core scripts and `false` at the end of them. This will store changes in special `core` tables that the user doesn't have "obvious" access to. This should prevent users from deleting core commands by accident.

Still need to write some help doc and tests before we're completely done.

PR #118, opened 2020-10-07
Tts impl

hjozwiak:
The work done thus far:
Added the tts crate to the project.
Added the tts flag to the flags structure.
Added the setting, defaulting to true.


LiquidityC:
I pulled this out and fixed the conflicts. There is a new branch in the repo now called `tts_impl` where we'll continue building the TTS functionality.

Issue #141, opened 2020-10-18
Add readline-style keybindings


xvxx:
It would be great if Blightmud supported more of the "standard" readline keybindings that shells, repls, etc support. If you are unfamiliar, check out the README of https://github.com/kkawakam/rustyline for an overview.

It's not as simple as including that crate since it does its own drawing and input handling, but Blightmud does have a good foundation with `step_word_right` and friends.

Would you be open to adding more built-in commands like this? 

Thanks, great project!

LiquidityC:
I replied to your PR but I'll add my comment here to for documentations sake. If you feel like digging into more of these bindings. In particular ctrl-delete/backspace and alt-delete/backspace you are very welcome.

Some of these bindings are highly sought after. They just haven't made it to the top of my list so far. As for special case handling as you mentioned in the PR. Those are fine. The scope is so isolated in this so any special handling in here is fine.

xvxx:
Sorry for the onslaught of noise, I've been looking for a simple Unix-y MUD client and just fell in love. Big fan of Lua, too, so for me the only thing that's really missing are all the obscure readline commands. 

I started working on some of the `backspace` stuff in #143. If you want to close this issue that's fine by me, I wasn't sure how you wanted to track meta conversations like this.

LiquidityC:
Don't mention it. The whole reason for doing OSS like this is that people can enjoy it and dive on in to the project. 

I think we're at a decent impasse in regards to what we can do with keybindings now so I'll close this one. Thanks for all the help!

I actually stole most of my CI scripts from your gopher project so you were contributing here before you knew it. 😊

Right now they are slightly borked but I don't think that's a human error.

xvxx:
> I actually stole most of my CI scripts from your gopher project so you were contributing here before you knew it. 😊

Amazing! Small world! 

I saw the CI was failing on macOS but I'm not sure why. I am using macOS and `cargo check` works fine for me, all the tests pass, etc. I wonder if the crate cache is somehow in a bad state.

LiquidityC:
Yeah, it will sometimes solve itself after a "cargo update" so yes. It's cache related but I don't really understand why. I'll give it 30 mins today.

PR #142, opened 2020-10-18
Adds "delete_word_left", "delete_word_right", and "delete_right"

xvxx:
This is related to #141. It adds three default bindings and their respective commands:

- `ctrl-d`: `delete_right`
- `alt-d`: `delete_word_right`
- `alt-h`: `delete_word_left`

The first two are standard Readline shortcuts, but the `delete_word_left` is non-standard. I made it `alt-h` to match `ctrl-h`, which is "move cursor back one."

The standard Readline keybinding for `delete_word_left` is `alt-delete`, but making that work in Blightmud will take a few changes to the key parsing. So I wanted to check and see if this direction is okay before adding more code. 

Specifically, the backspace/delete key sends `Alt('\u{07}')` which just gets `format!()`'d into `alt-`. So we would need to check for that character and represent it with something like `delete`. (Which probably wouldn't be bad because it would also make `ctrl-delete` usable as a binding, but it is the first special case with the ctrl/alt handling code as far as I can tell.) 

Anyway let me know what you think, or if you have any pointers, thanks!


LiquidityC:
Thanks. Looks good. If you want to take a swing at the backspace/delete combos you are more then welcome. All this keybindings stuff is very isolated so one or two special cases won't explode 😉

PR #143, opened 2020-10-19
Support for alt-escape and alt-backspace bindings.

xvxx:
This adds preliminary support for `alt-escape` and `alt-backspace` keybindings. In implementing this I learned that `ctrl-escape` and `ctrl-backspace` bindings don't work for various reasons:

- When I press the `ctrl` and `backspace` keys on my keyboard, Blightmud receives `ctrl-h`. This seems to be standard behavior, and `ctrl-h` is the standard readline shortcut for backspace, so I've added that binding. But I'm not sure `ctrl-backspace` is something we can let people bind to directly.
- When I press the `ctrl` and `escape` keys on my keyboard, Blightmud receives nothing. Same with the termion `keys` example, so I'm not sure if this is supported upstream.

I've also added support for the `delete` key, which is supposed to delete the character in front of the cursor (as opposed to behind it, as backspace does). On my keyboard this is `fn`+`backspace`.

Are you okay pulling this in now, or should I hunt down the `ctrl-escape` situation? My whole goal for this is getting `alt backspace` mapped to `delete_word_left`, so personally I am okay with these limitations, but let me know which direction to head. Thanks!


LiquidityC:
This looks solid. The unsolvable situations can remain like that. This is still a big improvement. I'll merge. Thanks for the PRs.

PR #144, opened 2020-10-19
Changes delete_word_* to share move_word_*'s behavior

xvxx:
I woke up this morning and thought, "*Dear Lord, what have I done?!*"

Instead of doing a bunch of questionable byte dancing, this implementation just re-uses `move_word_right()` and `move_word_left()`. I think it's less surprising for the user, too, since now all the "word" commands have the same behavior.

The tests pass unchanged so there shouldn't be any real difference, other than edge cases.

Original PR: #142

### DIFF
--- a/src/ui/command.rs
+++ b/src/ui/command.rs
@@ -163,31 +163,19 @@ impl CommandBuffer {
     }
 
     fn delete_word_right(&mut self) {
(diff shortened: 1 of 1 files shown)


LiquidityC:
Simple and effective 😊

Issue #13, opened 2020-05-18
Scroll output using scroll wheel


LiquidityC:
Scroll wheel has differen effects in different terminals. A common one seems to be "up/down arrow". I think termion is able to grab scroll wheel.

LiquidityC:
I can't figure out a good way of handling this. We're at the mercy of the terminal dev and type here. Moving out of project for the time being.

LiquidityC:
Don't think this one will ever really work out without a thorough investigation. Closing for now.

PR #145, opened 2020-10-20
Adds mouse wheel scrolling of output buffer.

xvxx:
This uses termion's `MouseTerminal` to implement #13. 

With this patch applied, you should be able to use the scroll wheel on your mouse to move the output buffer up and down (assuming there is more content than fits on a screen). 

If you don't have a mouse, then there should be no change.

I've only tested this on a Macbook, but I did test in kitty, Terminal.app, and Alacritty.

### DIFF
--- a/src/ui/command.rs
+++ b/src/ui/command.rs
@@ -254,6 +254,15 @@ impl CommandBuffer {
     }
 }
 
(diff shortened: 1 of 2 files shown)


LiquidityC (approved):
Looking good. Thanks

PR #152, opened 2020-10-22
Adds -v and --version flags

xvxx:
Also shows `[DEBUG]` unless Blightmud was compiled in `--release` mode.


LiquidityC (approved):
Hadn't thought about that. Thanks

Issue #148, opened 2020-10-22
Can't copy from terminal with 2.1.0, works in 2.0.1


dmytri:
I can't select and copy text from the terminal with 2.1.0, but it works in 2.0.1. Is this a bug or there a new setting?

LiquidityC:
We have noticed this. It's a side-effect from adding mouse scrolling. I'm not sure how to deal with it yet because I miss it too but I also enjoy having mouse scrolling.

dmytri:
Well if it's one or the other maybe a setting?

LiquidityC:
I'll do some digging when I get some time over. Preferably I'd like to have some sort of fall through for the events.

xvxx:
This is a bummer. As far as I can tell, the situation is that most terminals send arrow keys when the scroll wheel is used in an alternate buffer terminal app. The new scroll-wheel PR works by enabling mouse events, but there doesn’t seem to be any “fall through” option like there would be in JS.

I’ve been looking for terminal apps that 1) use the alternate buffer 2) allow scrolling with the mouse wheel and 3) allow native selecting of text. It does seem to be a rare combo.

In vim I get the worst of both worlds: my mouse wheel just seems to send arrow keys, and selecting text selects it in vim’s visual mode as opposed to my terminal’s default text selection. 

There is a lot online about tmux’s internal scrollback buffer. I think they hit all three criteria above by implementing their own text selection.

I am going to do some more research too. Maybe we can just add some simple text selection and mouse click detection, if we need to.

xvxx:
The other option would be to tell people that the mouse scrolling issue is a setting they can change in their terminal, and provide an FAQ. Assuming that’s true. Seems like at least iTerm2 surfaces this setting:

https://stackoverflow.com/questions/36594420/how-can-i-turn-off-scrolling-the-history-in-iterm2

LiquidityC:
`set mouse=a` or `set mouse=nvi` gives me good scrolling in vim using gnome-terminal. Sort of besides the point though. I'll do some digging but from what I can tell this will be a tricky one. Simple selection will probably be tricky.

Xandaros:
I use this in my config for mouse scrolling:
```lua
blight:bind("\u{1b}oa", function()
    blight:ui("scroll_up")
end)

blight:bind("\u{1b}ob", function()
    blight:ui("scroll_down")
end)
```
So switching to a mouse terminal is not required for mouse scrolling.
Although it would be required to get other mouse events...

[...]

PR #156, opened 2020-10-23
Optional mouse support

xvxx:
This addresses some of #148. 

Notably:

- Mouse support is now disabled by default
- You can enable mouse support by passing `-M` or `--mouse` when starting the app. (temporarily)

I think it's probably better for this setting to live in your `*.lua` config scripts so that you can make it permanent and keep it under version control, but I wanted to take a stab at this and get it in motion.

@LiquidityC Any guidance on what the config option should be?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -132,6 +132,7 @@ fn main() {
         "tts",
         "Use the TTS system when playing a MUD (for visually impaired users)",
     );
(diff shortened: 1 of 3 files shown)


LiquidityC:
This looks good for now. For the configuration I'm thinking it should reside in core. I'll drop an issue for that fix.

PR #165, opened 2020-10-30
Makes Lua's `print()` write to Blightmud's screen.

xvxx:
Just a tiny little thing. This makes Lua's `print()` write to Blightmud's screen.

Before:

```
> /lua print 'oh no'
(no output)
```

After:

```
> /lua print 'heyoo'
heyoo
```

### DIFF
--- a/resources/help/lua_blight.md
+++ b/resources/help/lua_blight.md
@@ -6,7 +6,7 @@ Blightmud's scripting engine.
 ##
 
 ***blight:output(str)***
-Prints output to the screen.
+Prints output to the screen. Also aliased as `print()`.
 - `str`  The string to output.
 ```lua
 -- Standard printing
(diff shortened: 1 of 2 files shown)


LiquidityC:
Awesome sauce! Just one note. Dropping a review comment.

LiquidityC (changes requested):
This is great. I've been thinking about fixing this. One note. Can you extend the documentation in `lua_blight.md`. Next to the `blight:output` documentation that this also works now?

Xandaros:
Hmm, do we really want to document `print`, though?
I feel like that's not on us, since it's a standard Lua function and the new behaviour is exactly what you would expect it to do

LiquidityC:
I was thinking in the lines of "Luas standard `print` function utilizes blight:output". Just as a short note. Otherwise people won't have the information available to them no big section on it.

xvxx:
@Xandaros I think it's worth mentioning because there are other Lua functions which don't yet work as you'd expect, like `io.write()` usually prints sans newline. Not sure if it's worth converting all of them or not. Maybe?

LiquidityC (approved):

LiquidityC:
We won't mess with `io.write` that might mess up file writing and we don't want that.

Issue #153, opened 2020-10-22
Function to remove single timer


Xandaros:
`blight:add_timer` and `blight:get_timer_ids` return timer IDs, but as far as I can tell, you can't actually do anything with those IDs currently.
There is `blight:clear_timers` to remove all timers, but no way to remove a single timer.

I suggest adding a `blight:remove_timer` function, which takes such a timer ID and removes the timer with the given ID.

PR #166, opened 2020-10-31
Adds blight:remove_timer() function. 

xvxx:
Addresses #153.

### DIFF
--- a/resources/help/lua_timers.md
+++ b/resources/help/lua_timers.md
@@ -21,19 +21,21 @@ end)
 
 ##
 
-***blight::remove_timer(timer_id)***
+***blight:remove_timer(timer_id)***
 
 - `timer_id` The id returned when creating a timer
 
 ```lua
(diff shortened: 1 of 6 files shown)


LiquidityC (changes requested):
The timer also needs to be removed from the timer thread. Look in `timer.rs`. Currently there's an event to clear all timers. Extending this with a method to drop a specific id should be simple enough.

Also, to be safe you might change `lua_script.rs:run_timed_function()` to handle if the requested callback is nil which could happen if you're unlucky with concurrency.

xvxx:
@LiquidityC I'm always unlucky with concurrency, so I added that nil check and am now removing the timer from the scheduler properly (in theory).

I updated the docs with a really simple test case, but am happy to try testing some more complicated timers too.

LiquidityC (approved):
Looking good now. Thanks

PR #169, opened 2020-10-31
Allows new settings

xvxx:
As far as I can tell, it's currently difficult to add new settings to existing installations of Blightmud because calling `settings.get()` with any key that does not appear in the `settings.toml` file will error, even if you've otherwise added the setting to all the right places in the code.

You can add a new setting to `Settings::default()`, but that's only used when there's no existing `settings.toml`.

This PR adds a list of known settings and checks against them before erroring, defaulting to `false` if one doesn't appear in `settings.toml`.

This doesn't address the "defaults" issue, however - any new settings added to `Settings::default()` will still be ignored when the TOML file already exists.


LiquidityC (approved):
Nice. Thanks

PR #170, opened 2020-10-31
Adds reloading of help files in debug mode.

xvxx:
Instead of loading all the help files on startup, this PR instead just sticks their paths into the `HelpHandler::files` HashMap and reads them from disk every time they're requested in debug mode.

The idea here is you can edit help files and see the changes you make instantly during development.

Release mode should still work the same.


LiquidityC (approved):
Very nice

PR #171, opened 2020-10-31
Adds setting to enable mouse support

xvxx:
This follows on from #156 and depends on #169. 

The `--mouse` CLI flag isn't great because you have to type it every time, or alias your `blightmud` binary, if you want to use your mouse.

Settings, however, can be kept in version control and don't require any special startup incantation.

This PR removes the `--mouse` and `-M` CLI flags and replaces them with a `mouse_enabled` setting.

I mentioned it in the help file, but changing this setting requires a restart. 

_Here's the diff of just the changes in this PR, without showing the settings changes from #169: https://github.com/xvxx/Blightmud/compare/allow-new-settings...adds-mouse-setting_


LiquidityC (approved):
Well done

LiquidityC:
Ok, some conflicts based on the earlier PR I guess.

PR #177, opened 2020-11-02
Persistent command history

xvxx:
This PR adds persistent command history and a `save_history` setting that controls it. The setting is off by default.

When it's turned on, Blightmud will save the last 100 commands you typed in `data/history.ron` when it shuts down and load from that file when it starts up. 

I had to shuffle some things around to make this work (notably loading `Settings` before building the `Session`), but there shouldn't be any behavior changes.

Two things to note:

1. Your MUD password(s) will be saved along with everything else. Is there any way around this?

2.  When you disable this setting after using it, your `history.ron` will still live on disk. However, if you then decide to re-enable the setting after restarting the app, `history.ron` will get overwritten with the history of your current session. I think that's okay, but wanted to note it.

It might also be nice to also have a setting for how many entries to save, if someone feels like 100 is too few.

Not really sure about the names I picked. Should `save_history` be `persistent_history` or something? Any advice or guidance would be most welcome.


LiquidityC:
Nice job. I'm ok with the decisions you made in regards to naming etc. Last 100 commands seems reasonable too. I think it's a very long command if hitting up 100+ times is faster then retyping it and I don't know many muds who require such long commands on the regular.

Issue #183, opened 2020-11-03
os.sleep() (or something...)


xvxx:
A MUD I play on tends to write directions like this:

    NEWBIE AREA  6n, 2w
    MID AREA     5s, 2w, Request Transit, Enter Module, 5s

So I added a little alias in Lua that lets me type `go 5n` to send `blight:send("n")` five times.

But, when doing lots of moves (like `go 15n`) I get rate limited - "You cannot move so fast!"

Lua doesn't come with `os.sleep()` and using `core:exec("sleep 1")` blocks the whole UI, so it'd be nice if there were a way to sleep between commands so I'm not overloading the server.

Xandaros:
I don't think this is a good idea. What happens if you are sleeping and an event happens, like a trigger or a timer firing? Does it happen after the sleep finishes? Does it stop the code with the sleep?

Instead, I think you should just use a timer. You can store the list of moves in a table and when a timer fires, you make a move. Something like this:
```lua
moves = {}
blight:add_timer(1, 0, function()
    for _, move in ipairs(moves) do
        blight:send(table.remove(moves, 1))
    end
end)
```
You can then do something like `moves = ["n", "n", "n", "w"}` to set it going.

Alternatively, if you prefer to use an actual sleep, you can run it in a coroutine and use a timer to resume it:
```lua
tasks = {}
blight:add_timer(0.2, 0, function()
    for _, task in ipairs(tasks) do
        if os.time() > task[1] then
            local delay = coroutine.resume(task[2])
            task[1] = os.time() + delay
        end
    end
end)
```
You can then use it like this:
```lua
tasks[#tasks + 1] = {0, coroutine.create(function()
    for i = 1, 5 do
        blight:send("n")
        coroutine.yield(1)
    end
end)}
```
Untested, but you get the general idea.

xvxx:
> What happens if you are sleeping and an event happens, like a trigger or a timer firing? Does it happen after the sleep finishes? Does it stop the code with the sleep?

Usually the `os.sleep()` is implemented as a coroutine in Lua (basically a timer behind the scenes), so everything else would happen normally. It wouldn't block other threads or coroutines. 

I haven't looked into how this would be implemented in Blightmud, but since the app is multithreaded it would probably be a similar principle.

Xandaros:
Yeah, but they have a different architecture.

I think what I described is still the best option, but I can make a scheduler that is a bit more robust for inclusion in Blightmud.
[...]

PR #185, opened 2020-11-05
Add tasks Lua library

Xandaros:
Adds a task scheduler

Documentation with some examples is included.

Also allows you to pause execution for a given number of seconds (See #183)

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -9,10 +9,10 @@ use timer::{Guard, MessageTimer};
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum TimerEvent {
-    Create(Duration, Option<u32>, u32),
+    Create(Duration, Option<u32>, u32, bool),
     Trigger(u32),
     Remove(u32),
-    Clear,
+    Clear(bool),
     Quit,
 }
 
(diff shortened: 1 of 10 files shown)


LiquidityC (changes requested):
This looks good. Just one problem. The timers don't have a core table. This means that a user can access and remove the task timer (by accident). There's a whole core mode system in blight that deals with this but it's not extended for timers. Can you include timers in this system to?

Xandaros:
Done. This makes some of the changed I made to the timer tests irrelevant, but I think they are a good idea anyway, so I'm keeping them.

src/timer.rs
```diff
@@ -24,18 +24,25 @@ struct Job {
 struct Schedule {
     main_thread_writer: Sender<Event>,
     jobs: HashMap<u32, Job>,
+    core_jobs: HashMap<u32, Job>,
```
LiquidityC, line 27: There is no way of clearing these jobs. Every time a user calls `blight:reset()` these timers won't be cleared but a new timer will be added when `tasks.lua` loads.

For Èvent::ResetScript` in `main.rs` a method that clears all the core timers needs to be called as well. Otherwise one might queue up a lot of timers. In my use case this will become a lot of timers.

LiquidityC (changes requested):

LiquidityC (approved):

PR #188, opened 2020-11-05
Adds regular expressions to lua

### LINKED ISSUE #176: Add Lua bindings for rust regex
Lua has pattern matching, but... it leaves a lot to be desired.
Since we are already using the `regex` crate, maybe adding bindings to that might be a good idea?

Users can still choose whether they use Lua's pattern matching or rust's regex, but it allows access to a more powerful regex library.

Especially since rust's regex is already used for triggers and such, but there is currently no way to utilise it directly.

LiquidityC:
Fixes #176

### DIFF
--- a/resources/help/regex.md
+++ b/resources/help/regex.md
@@ -1,0 +1,61 @@
+# Regular expressions
+
+Blightmuds lua script offers access to powerful regular expressions. These can
+be created and used to match and replace content in strings.
+
+##
+
+***regex.new(pattern)***
+Creates a new regular expression.
+
+##
+
+***regex:test(string)***
+Checks if a string matches the regular expression.
+Returns `true` or `false`.
+
+```lua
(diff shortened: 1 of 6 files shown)


xvxx:
This is awesome. I noticed the triggers weren't using Lua's PEG syntax, it's real nice to have actual regexs available from Lua.

resources/help/regex.md
```diff
@@ -0,0 +1,61 @@
...
+
+##
+
+***regex:test(string)***
+Checks if a string matches the regular expression.
+Returns `true` or `false`.
+
+```lua
+local re = regex.new("^This is a \\w+ line$")
+assert(re.test("This is a good line"))
```
xvxx, line 19: These should be `re:test()` I believe, not `re.test()`.

LiquidityC: Good spot. I'll have to patch that :smile:

PR #190, opened 2020-11-05
Confirm quit setting

xvxx:
I love how fast Blightmud starts and quits, but I may or may not have accidentally hit `ctrl-c` in the midst of some intense text-based action the other day and had to quickly scramble to reconnect.

This adds a `confirm_quit` setting that, when enabled, will ask you to hit `ctrl-c` to confirm that you really want to quit the app:

```
[**] Confirm quit with ctrl-c
```

You'll get prompted whether you try to quit with `ctrl-c` or `/quit`, but in both cases you must hit `ctrl-c` to confirm.

I think it's nice because you can still hit `ctrl-c` twice in a row to quit really quickly, but you've also got some bumper lanes if you want them.


src/model/settings.rs
```diff
@@ -13,7 +13,14 @@ pub const LOGGING_ENABLED: &str = "logging_enabled";
...
 pub const MOUSE_ENABLED: &str = "mouse_enabled";
 pub const SAVE_HISTORY: &str = "save_history";
-pub const SETTINGS: [&str; 4] = [LOGGING_ENABLED, TTS_ENABLED, MOUSE_ENABLED, SAVE_HISTORY];
+pub const CONFIRM_QUIT: &str = "confirm_quit";
+pub const SETTINGS: [&str; 5] = [
+    LOGGING_ENABLED,
+    TTS_ENABLED,
+    MOUSE_ENABLED,
+    SAVE_HISTORY,
+    CONFIRM_QUIT,
```
LiquidityC, line 22: Can you default this to true a few lines below please? I think most people will want it.

LiquidityC (changes requested):

src/model/settings.rs
xvxx: Changed in 3b2a80a150bd7b4296247e3862310b46c0ed1599

LiquidityC: Thanks

LiquidityC:
This is a good addition. I've both accidentally hit `ctrl-c` and typed `/q` at one point or another. Thanks

PR #191, opened 2020-11-05
Adds 'home' key shortcut to scroll to top

xvxx:
Completes the holy quadrinity of home / end / page up / page down. 👼


LiquidityC (approved):

LiquidityC (changes requested):
In `bindings.md` there is a copy of the current binding setup. Could you please replace that with the *new* current version. I'm guessing you are on vim. So some deletes and then a `:read resources/lua/bindings.lua` would get it right in there.

xvxx:
Just updated :+1: Looks like `home`/`end`/ and friends aren't in there because they're defined in Rust tho.

LiquidityC:
Yes I noticed. I'm sorry, I thought they were in there. Thanks for adding this though. It was still needed.

PR #192, opened 2020-11-05
Removes `terminate` in favor of Event::Quit

xvxx:
Something I noticed when messing with `ctrl-c` is that if we can make everything send `Event::Quit` to the main thread, we probably don't need to use the session's `terminate` boolean anymore, unless there are future plans for it or something I'm overlooking. All the child threads will get killed when the main thread exits, unless we `join()` on them somewhere, which we're not at this point. As a bonus it also knocks off a level of indentation in the big loop in `main.rs`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,10 +2,7 @@ use lazy_static::lazy_static;
 use libtelnet_rs::events::TelnetEvents;
 use log::{error, info};
 use std::path::PathBuf;
-use std::sync::{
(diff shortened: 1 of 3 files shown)


LiquidityC (approved):
Yes, you are right it is no longer needed. Thanks

LiquidityC:
Just some conflicts then I'll merge

xvxx:
Just updated, should merge cleanly now.

PR #201, opened 2020-11-07
Realtime logging

xvxx:
Not a big change, but this means MUD log files will get written to as events occur rather than when you quit the app.

After @LiquidityC blew my mind with https://github.com/LiquidityC/Blightmud/issues/195#issuecomment-723343842 I realized that I could be really lazy and avoid writing any Lua at all if Blightmud wrote to its logfile in realtime. As in, you can just do a `tail -f | grep` on the logfile for simple patterns. Or write entire programs that watch the logfile looking for patterns and doing stuff when it finds them...

### DIFF
--- a/src/io/logger.rs
+++ b/src/io/logger.rs
@@ -11,16 +11,8 @@ pub struct Logger {
     file: Option<BufWriter<StripWriter<File>>>,
 }
 
-#[cfg(not(test))]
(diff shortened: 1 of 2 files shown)


LiquidityC (changes requested):
This looks good. But it's not building.

PR #202, opened 2020-11-08
Human friendly settings and servers

xvxx:
Something I want to propose while [Version 3.0](https://github.com/LiquidityC/Blightmud/issues/194) is on the table is making the settings and servers files a bit more human-friendly so they are easy to keep in version control and easy to edit with an `$EDITOR`. 

These are breaking changes right now, but I was thinking we could have an `/upgrade 3.0` command or something to automatically transfer "old" settings/servers to their new locations. Open to suggestions of course. 

Specifically I'd propose moving `settings.ron` and `servers.ron` to `CONFIG_DIR` so they live alongside your `*.lua` files, having the `ron` crate pretty-print them so they are easier to edit, and making the tls setting just `bool` instead of `Option<bool>`.

My thinking here is that you probably don't want to keep your log files under version control and share them between computers, but you will definitely want to keep your settings, servers, and Lua scripts in version control and share them. And if you do want to share your log files, you still can, of course. 

Before:
```
$ cat ~/.local/share/blightmud/data/servers.ron
{"lotr":(host:"lotr-mud.com",port:5555,tls:Some(false)),"starwars":(host:"starwars.mud.com",port:1212,tls:Some(false)),"mymud":(host:"mud",port:6789,tls:Some(false))}
```

After:

```rust
$ cat ~/.config/blightmud/servers.ron
{
    "mymud": (
        host: "mud",
        port: 6789,
        tls: false,
    ),
    "starwars": (
        host: "starwars.mud.com",
        port: 1212,
        tls: false,
    ),
    "lotr": (
        host: "lotr-mud.com",
        port: 5555,
        tls: false,
    ),
}
```

Before:

```rust
$ cat ~/.local/share/blightmud/data/settings.ron
(settings:{"confirm_quit":false,"mouse_enabled":true,"logging_enabled":false,"tts_enabled":true,"save_history":true})
```

After:

```rust
$ cat ~/.config/blightmud/settings.ron
{
    "logging_enabled": false,
    "confirm_quit": false,
    "save_history": false,
    "tts_enabled": true,
    "mouse_enabled": true,
}
```


LiquidityC (changes requested):
I can get onboard with this. But I'd like to avoid user interaction.
Create a `patch.rs` or `updates.rs` in `tools` mod and have it provide a function that moves these files if they haven't been moved to the new place. Call this function when blightmud starts.
You might even lock it down to only run if the running version is `<= 3.0.0`
If it's possible it could even print something helpful if it does make any files changes.

Update the helpfiles regarding settings and servers to mention where these files exist. Note especially that edits to them while blightmud is running will be overwritten when blightmud shuts down.

We also need to update the loading of these files. If the file isn't correctly formatted blightmud will ignore it and (probably) overwrite it on exit. This logic has to be updated. First we need to tell the user that the file is no good. Then we need to remember that the file was no good and not overwrite it on exit.

xvxx:
Okay, I've made some progress on this. I'm going off of this list, as per @LiquidityC's comment:

- [x] Create `patch.rs` in `tools` mod 
- [x] Have it provide a function that moves these files if they haven't been moved to the new place.
- [x] Call this function when blightmud starts.
- [x] You might even lock it down to only run if the running version is >= 3.0.0
- [x] If it's possible it could even print something helpful if it does make any files changes.
- [ ] Update the helpfiles regarding settings and servers to mention where these files exist. 
- [ ] If the file isn't correctly formatted, we need to tell the user that the file is no good.
- [ ] bonus: make sure tts_settings.ron is migrated, too.

It only runs in release mode when version is >= 3.0.0. 

If the new settings file already exists, it won't overwrite it. Instead you'll get an error:

```
[!!] Can't migrate settings.ron:
[!!] settings.ron already exists at "/Users/xvxx/.config/blightmud/settings.ron"
```

Otherwise, you'll get a message that the migration succeeded:

```
[**] Migrated servers.ron from "/Users/xvxx/.local/share/blightmud/data/servers.ron" to "/Users/xvxx/.config/blightmud/servers.ron"
[**] Migrated settings.ron from "/Users/xvxx/.local/share/blightmud/data/servers.ron" to "/Users/xvxx/.local/share/blightmud/settings.ron"
```

Nothing will print if there's nothing to be done.

xvxx:
I got everything on the list except the version constraint - since the paths are changed, there can be problems when the migrator doesn't run and the old file formats are saved to the new paths. So I think this should only get merged into the branch that becomes 3.0 to avoid any issues.

I also added some tests, they're a bit weird because they hit the file system and therefor can't be run in parallel. There is some setup/cleanup involved as a result. But they helped me find a few bugs, so I think it was a good exercise.

Let me know what you'd like me to change/remove/add thanks!

LiquidityC:
Very nice work! I'll merge this later when I have a chance to run it locally. Thanks for all your effort!

PR #203, opened 2020-11-08
Adds test_lua!(), assert_lua!(), and other helpers

xvxx:
I wanted to see if there's any interest in moving this direction while we're discussing testing Lua code. I've found Rust macros to be super helpful in cleaning up tests and providing little helpers to reduce boilerplate, so I added a few to `line.rs` as a little experiment. They're all generic, so we could pull them out into `test_helper.rs` mod or something if we want to move forward.

But in a nutshell, I've added a macro, `test_lua!()`, that initializes the Lua state by calling `get_lua()` and then adds a few Lua-specific macros to each test function: 

- `assert_lua_string!(code, expected)`
- `assert_lua_bool!(code, expected)`
- `assert_lua!(TYPE, code, expected)`
- `let val: TYPE = global!("key");`

So this:

```rust
let state = get_lua();
assert!(!state
    .context(|ctx| -> bool { ctx.load("return test_line:prompt()").call(()).unwrap() }));
```

becomes this:

```rust
test_lua!();
assert_lua_bool!("test_line:prompt()", false);
```

And this:

```rust
let state = get_lua();
assert_eq!(
    state
        .context(|ctx| -> String { ctx.load("return test_line:line()").call(()).unwrap() }),
            "A testing line"
);
```
becomes this:

```rust
test_lua!();
assert_lua_string!("test_line:line()", "A testing line");
```

And finally, this:

```rust
let state = get_lua();
let line = state.context(|ctx| -> Line { ctx.globals().get("test_line").unwrap() });
assert!(line.inner.flags.tts_gag);
```

becomes:

```rust
test_lua!();
let line: Line = global!("test_line");
assert!(line.inner.flags.tts_gag);
```

There's also a generic `assert_lua!()` for other types, which the others wrap:

```rust
test_lua!();
assert_lua!(usize, "123", 123);
```

The `test_lua!()` exists so we don't have to pass around the `state`, but we can surface that if it's convenient.

### DIFF
--- a/src/lua/line.rs
+++ b/src/lua/line.rs
@@ -70,6 +70,7 @@ impl UserData for Line {
 }
 
 #[cfg(test)]
+#[allow(unused_macros)]
 mod test_lua_line {
     use rlua::Lua;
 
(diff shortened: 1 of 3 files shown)


LiquidityC (changes requested):
This looks great. Just some requests. Could you put it in a test_help.rs file directly. I'll definitely use this a lot during coming changes.

Could we also add a macro `set_global!(UserData)` and instead use this to set the `test_line` global. This way it can be used easily for existing and old lua modules. Also it will allow for more dynamic test functions. I used one single global for line.rs out of laziness.

xvxx:
Sounds good. I moved it all into `src/lua/test_help.rs` and added `set_global!(key, value)`. I also added syntax to `test_lua!()` to make setting initial globals easier, so this:

```rust
test_lua!();
set_global!("test_line", Line::from(mLine::from("\x1b[31mA testing line\x1b[0m")));
```

Is the same as this:

```rust
test_lua!("test_line" => Line::from(mLine::from("\x1b[31mA testing line\x1b[0m")));
```

You can mix and match, `test_lua!()` just calls `set_global!()` behind the scenes now.

PR #212, opened 2020-11-13
Move triggers to Lua

### LINKED ISSUE #205: Move trigger system to pure lua
As described in #194 move all of the trigger handling to a clean lua implementation. Regex and registering for mud output lines is available already.

All existing methods must be moved to the new `trigger` module. Some names may be changed. Especially if they have the word 'trigger' in the name. Eg. `blight:add_trigger() -> trigger:add()`.

Update doc for all changes made.

This can be started as soon as #197 is merged.

Xandaros:
Closes #205 

Still WIP:
- [x] Implement trigger system in Lua
- [x] Change tests to use new trigger system
- [x] Update documentation


Xandaros:
This is now done. #214 should probably be fixed before this gets merged, though

resources/lua/triggers.lua
```diff
@@ -0,0 +1,203 @@
...
+end
+
+mud:add_output_listener(function(line)
+    for _, group in pairs(system_trigger_groups) do
+        group:checkLine(line)
+    end
+    for _, group in pairs(user_trigger_groups) do
+        group:checkLine(line)
+    end
+    return line
```
LiquidityC, line 200: You should differentiate between prompt and regular output here. Prompt triggers should only match prompt output. This logic is ready and available in dev now. Eg. `Line:prompt() => bool`

src/lua/lua_script.rs
```diff
@@ -84,6 +84,10 @@ fn create_default_lua_state(
             .load(include_str!("../../resources/lua/tasks.lua"))
             .call::<_, rlua::Value>(())?;
         globals.set("tasks", lua_tasks)?;
+        let lua_triggers = ctx
+            .load(include_str!("../../resources/lua/triggers.lua"))
+            .call::<_, rlua::Value>(())?;
+        globals.set("triggers", lua_triggers)?;
```
LiquidityC, line 90: Id prefer if the module is named `trigger`. For me it feels nicer with `trigger.add()` and `trigger.remove()` etc.

resources/lua/triggers.lua
LiquidityC: Well, perhaps not here. But you get it :wink:

LiquidityC (changes requested):

src/lua/lua_script.rs
Xandaros: Hmm, in that case we should probably rename `tasks` to `task`, as well

resources/lua/triggers.lua
LiquidityC: Saw your comment. Ignore this

src/lua/lua_script.rs
LiquidityC: On an additional note. We don't need to require this one through rust do we? We could just require it in one of the lua files. Eg. Add one called systems which loads them?

LiquidityC: Or perhaps we do to make sure they get baked into memory...

Xandaros: Yeah, currently they are part of the executable.
The minecraft mods that run Lua code tend to not place their libraries in `_G` and instead require you to always `require` them. I'm not sure we can do that easily in Blightmud without going away from the single executable, which I honestly quite like.

It's certainly possible - worst case, we monkey-patch `require` - but that seems a bit... drastic.

LiquidityC: Nah. We're sticking with single executable. We will do it like this for our mods and use the other way for plugins.

src/lua/lua_script.rs
```diff
@@ -84,6 +84,10 @@ fn create_default_lua_state(
             .load(include_str!("../../resources/lua/tasks.lua"))
             .call::<_, rlua::Value>(())?;
         globals.set("tasks", lua_tasks)?;
+        let lua_triggers = ctx
+            .load(include_str!("../../resources/lua/trigger.lua"))
+            .call::<_, rlua::Value>(())?;
+        globals.set("trigger", lua_triggers)?;
```
LiquidityC, line 90: The doc doesn't match this naming now. 😋

LiquidityC (changes requested):

src/lua/lua_script.rs
Xandaros: oops

LiquidityC:
Awesome work! Thanks

Issue #208, opened 2020-11-13
Move logging functions from blight to new log module


LiquidityC:
As described in #194. Move all the logging stuff from `blight` to a new `log` module.

Issue #194, opened 2020-11-06
Version 3.0


LiquidityC:
# Version 3.0

This is the main task wrapping all the changes in store for the next major release. It won't add much new functionality but it will open up blightmud a lot more for future plugins and similar.
Issues should be created for every point on the list. If you want to work on something but can't find the issue then create it.

## Breaking changes

- [x] Introduce a Lua object `Line` that mimics the current `Line.rs` but in lua. `lua::Line::from(model::Line)` ***blocker, needs doing first***
- [x] Send all mud output lines to lua as `Line` objects. I'm thinking `mud:add_output_listener(callback)`
- [x] Send all user input to lua as `Line` objects. `mud:add_input_listener(callback)`
- [x] Break out the trigger functions from `blight.rs` and re-build it in pure lua using the new `Regex` module. (new `trigger` module)
- [x] Break out the alias functions from `blight.rs` and re-build it in pure lua using the new `Regex` module. (new `alias` module)
- [x] Move the timer related code out of `blight` and put it in it's own module `timers`. This will have to remain a lua/rust hybrid because of the timer thread.
- [x] Move the logging related stuff from `blight` into a new `logging` module.
- [x] Move all the mud related stuff **send, send_bytes, mud_output, disconnect, connect, on_(dis)connect** to the new `mud` module mentioned above.
- [x] Move script related functions (**load, reset**) from `blight` to a new `script` module.
- [x] Move store functions from blight to a new `store` module. Also move the session store from `core` to this new module and give the methods names based on storage type.
- [x] Move settings aliases out of rust into a `settings` module and setup the aliases in lua instead.

Note, the above should leave `blight` with only one function: **output**. I think that will still belong. Because it's about doing stuff with blightmuds screen and display. Unrelated to any other system. Also people will most likely only use print now.

## Think about
Here are some additional things that should be kept in mind if you're helping out with the next big version.

### Naming
Some functions in `blight` currently have names that pertain to their responsibility. When creating a more logical module system some of these names will be stupid. If you feel that way then change the name. Eg. `blight:start_logging()` **=>** `logging:start()`

### Lua code
Some of these changes will break existing scripts we have in place. `gmcp`, `msdp`, `macros` etc. Make sure you check these. Simple find/replace will most likely suffice.

### Testing
[...]

PR #231, opened 2020-12-05
Move logging functions to separate module (#208)

singularpoint:
* start_log and stop_log functions moved to separate log module

* Actualizes help files about scripting and changes

* Fixes problem with duplication of messages about logging stop

As described in #194


LiquidityC (approved):
Looks good

Issue #232, opened 2020-12-06
Move store/read functions from blight/core to a new store module.


LiquidityC:
Relates to main task #194 

Currently there are two methods for storing.

- `blight:store()` and `blight:read()` which writes and reads to/from disk.
- `core:store()` and `core:read()` which writes and reads to/from memory (eg. doesn't persist between restarts).

New methods will be:
- `store.disk_write(key, val)` and `store.disk_read(key)`
- `store.session_write(key, val)` and `store.session_read(key)`

Xandaros:
Omitting `val` is the same as specifying `nil` for it. Personally, I would expect that to remove the key from the store instead of reading it.

LiquidityC:
So you'd rather have `store.session_write()` and `store.session_read()` ? Is that more conforming to lua conventions?

Xandaros:
I don't think there is anything particular to Lua about this, but yes, that is what I would do.
With `store.session_write(key, nil)` removing the key entirely.
Similar to how setting a value to `nil` in a table removes it.

LiquidityC:
In rust we can differentiate between `None`, `rlua::Nil` and `rlua::Value` though. So supplying `nil` could still delete the key while supplying nothing can just read it. That's a perk in rust though and not something that can be replicated in other lua implementations.

LiquidityC:
I'll update the main comment. I think you might be on to the right idea just for general clarity.

singularpoint:
Do we really need not-disk storage for session-wide values? We can just use a global Lua table for this.

LiquidityC:
Yes. This session storage persists through a 'script.reset()'. I use it a lot in my scripts to keep game session data available even if I reset and reload my scripts. It's really handy if you are working on scripts and reloading a lot but still want game session information to remain in memory. Combined with the json module you can store advanced tables.

PR #237, opened 2020-12-26
Move storage functions to separate module (#232)

singularpoint:
* Both disk and in-memory persistent storage function moved to store.rs

* After script.reset() store module survives, and core module don't

* Documentation updates


LiquidityC (approved):
Nice work!

Issue #238, opened 2020-12-26
Move settings aliases to lua


LiquidityC:
As described in #194 

Currently there are a bunch of `/set` and `/settings` commands that could be moved to a lua module. The aliases could then be created in lua instead.

The module should allow scripts to get and set settings as well as retrieve all available settings.

singularpoint:
I can take the issue.

LiquidityC:
@singularpoint If you got the time you are welcome to start this one. I have limited time until next year. I'll assign you. Just comment if you'd rather not :wink:

LiquidityC:
@xvxx You have any input on this issue? If I'm not mistaken you presented the idea that created it :smile: 
Or did I sum it up just about right in the details above?

PR #243, opened 2020-12-29
Move settings aliases to lua (#238)

singularpoint:
* remove commands related to settings
* remove core events of displaying settings
* add lua module to handle lua commands
* add scripting docs related to settings
* add setting update core event to handle
  history_saving atomic value correctly
* add lua aliases /settings and /set


resources/lua/macros.lua
```diff
@@ -60,3 +60,40 @@ end)
...
+		else
+			value = matches[3] == "on"
+			settings.set(key, value)
+		end
+		local key_format = cformat("<yellow>%s<reset>", key)
+		local value_format
+		if value then
+			value_format = cformat("<bwhite>on<reset>")
+		else
+			value_format = cformat("<bblack>off<reset>")
```
LiquidityC, line 95: Don't use "bblack" for off here. That'll be the background color on some terminals (like mine :grin:). I'd prefer bgreen for on and bred for off.

resources/lua/macros.lua
```diff
@@ -60,3 +60,40 @@ end)
...
+		end
+		blight.output(cformat("%s => %s", key_format, value_format))
```
LiquidityC, line 97: Prefix this string with `[**] ` to conform with the old functionality. Also it's the blightmud standard for printing "info".

```lua
blight.output(cformat("[**] %s => %s", key_format, value_format))
```
This hasn't become the standard for all the internal lua generated output. But I'd like to move in that direction instead.

LiquidityC (changes requested):
Nice work. Just a few comments on UX/Visuals

resources/lua/macros.lua
singularpoint: Done in the updated commit.

resources/lua/macros.lua
singularpoint: Done in the updated commit.

LiquidityC (approved):

Issue #242, opened 2020-12-28
history saving doesn't seem to work


singularpoint:
### Input history is never saved

This happens because in current implementation history should be saved in the input thread after the stdin event loop, but the thread never leave this loop. As a result history never saved no matter what the value of save_history parameter is.

PR #244, opened 2020-12-30
Fixes history saving

LiquidityC:
Fixes: #242

### DIFF
--- a/src/ui/command.rs
+++ b/src/ui/command.rs
@@ -286,12 +286,19 @@ fn parse_key_event(
     buffer: &mut CommandBuffer,
     writer: &Sender<Event>,
     tts_ctrl: &mut Arc<Mutex<TTSController>>,
+    save_history: bool,
 ) {
     match key {
         Key::Char('\n') => {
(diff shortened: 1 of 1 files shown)


LiquidityC:
This is sort of a quick and dirty fix.
The downsides:
- Save history is now a "next restart" setting. (I can live with this)
- If you have "confirm quit" enabled the history will save twice. (I can live with this too)
- The input handling is getting sort of messy (this should be fixed by moving more of the commands to lua implementations. (Next in line is servers in my opinion).

src/ui/command.rs
```diff
@@ -286,12 +286,19 @@ fn parse_key_event(
     buffer: &mut CommandBuffer,
     writer: &Sender<Event>,
     tts_ctrl: &mut Arc<Mutex<TTSController>>,
+    save_history: bool,
```
singularpoint, line 289: save_history is Arc<AtomicBool> in Session object, so we can safely share it among threads. I think in future we can refactor everything to share full Settings object as sort of context for many threads and components.

singularpoint (approved):

PR #267, opened 2021-02-13
audio crash fix

LiquidityC:
- Attempts to prevent bad error handling in rodio from crashing blightmud
- Changes rodio dep to patched version re-exporting cpal
- Hides audio init behind feature

Since rodio seems slightly unstable I'm placing the audio feature behind an opt-in feature
for the time being. We'll have to decide how to continue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,7 +185,6 @@ version = "3.1.0"
 dependencies = [
  "anyhow",
  "chrono",
- "cpal",
  "curl",
  "flate2",
  "getopts",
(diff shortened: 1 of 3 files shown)


LiquidityC:
Pending this RustAudio/rodio#350
If above PR is merged we should be able to scrap this PR since the error causing a problem should no longer crash the app.

LiquidityC:
Rodio has pushed a new version of their crate now which should fix our problem. Dropping this PR now.

PR #282, opened 2021-03-18
Switches default plugin branch from `master` -> `main`

LiquidityC:


LiquidityC:
We will hold off on this until git ha released their change of the default branch name.

LiquidityC:
Closing this. A dynamic solution has been made that will fallback on main if master doesn't exist.

PR #323, opened 2021-06-07
Make mlua branch compile

Xandaros:
Tests don't work yet

### DIFF
--- a/src/lua/lua_script.rs
+++ b/src/lua/lua_script.rs
@@ -23,7 +23,7 @@ fn create_default_lua_state(
     dimensions: (u16, u16),
     store: Option<Store>,
 ) -> Lua {
-    let state = unsafe { Lua::new_with_debug() };
+    let state = unsafe { Lua::unsafe_new() };
 
     let backend = Backend::new(writer.clone());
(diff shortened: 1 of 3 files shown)


src/lua/lua_script.rs
```diff
@@ -141,31 +148,23 @@ impl LuaScript {
...
-        self.state
-            .context(|ctx| -> LuaResult<Vec<Line>> {
-                let blight_aud: AnyUserData = ctx.globals().get("blight")?;
-                let mut blight = blight_aud.borrow_mut::<Blight>()?;
-                let lines = blight.get_output_lines();
-                Ok(lines)
-            })
-            .unwrap()
+        let blight_aud: AnyUserData = self.state.globals().get("blight").unwrap();
+        let mut blight = blight_aud.borrow_mut::<Blight>().unwrap();
```
LiquidityC, line 157: I'd prefer `.expect("Some message about bad 'blight' object")` since we are trying to reduce the unwrap count. Expects makes it more clear that this a valid hard-crash case.

LiquidityC (changes requested):
Tests also went belly up.

Issue #335, opened 2021-06-20
Custom tab completion


Xandaros:
It would be nice if it were possible to override the default completion behaviour.

I'm thinking of something like this:
`blight.on_tab_complete(callback: fun(current_input: string): string | nil)`

Multiple handlers may be registered. Execution order is undefined. The first handler that returns a string will cause that string to be appended to the input and prevent further handlers from being executed. Handlers who do not wish to handle completion should return `nil`. An empty string is a valid completion.

This could be useful in many cases:
- Commands could be auto-completed, like `/qui<tab>` turns into `/quit` or `/reconn` turning into `/reconnect`
- Command arguments (Full `/lua` completion, anyone? Probably not. But maybe arguments to `/connect` and the like)
- Basically all of the above, but with mud-specific commands. Particularly if you are in a mode with limited commands (Like Starmourn's hacking, for example)

LiquidityC:
Sounds like a reasonable improvement. Especially if people want to make mud specific plugins.

LiquidityC:
I don't recall if the current completion returns a full word or just the completion suffix. Either way this addition will have to match that behavior.

Xandaros:
Yeah, I forgot to mention that the current completion behaviour will run as fallback if all handlers returned `nil`.

Though, to be fair, that is pretty obvious :P

LiquidityC:
I took a look at this. The existing completion operates by returning a list of completions. Consecutive `<tab>` presses will step through this returned list without calling the completion function again. Most typing actions will clear the list and future `<tab>` presses will call the completion function again.

The custom completions will have to mimic this behavior by returning lists of completions. This would allow multiple custom completions to work in tandem by just concatenating the lists of completions that they return including the core implementation.

The order of the available completions will be based on the order of completion execution and the only guarantee we give is that the core completions will be last in that list.

To me this sounds like a better implementation. More in line with how eg. `bash_completion` works.

The one downside would be if a user has lots of custom completers which could potentially yield a very long list of completion options. But I don't really see this as a major problem. Same thing happens in most completion implementations if you start hitting tab after just typing one character.

Thoughts @Xandaros ?

PR #338, opened 2021-06-21
Introduces custom tab completion callbacks

LiquidityC:
Fixes: #335

### DIFF
--- a/resources/help/lua_blight.md
+++ b/resources/help/lua_blight.md
@@ -69,6 +69,36 @@ Searches for a string forward from current position
 
 ##
 
+***blight.on_complete(callback: function(input: string) -> [string] | nil)***
+Allows users to insert custom tab completion logic into Blightmud
+
+- `callback`    The function that gets called on a `complete` event. 
+
+The callback function should return a list of completions or `nil` if no
+completions were found. The completions should be complete words (prefix
+included) and not just the suffix or completion part of the word.
+
+You may register multiple completion callbacks, the execution order of these is
+undefined however the default completion inside Blightmud will always be
+executed after custom completion functions.
+
+All completions from custom functions and default completions will be
+concatenated into a list with duplicates removed (order preserved). Subsequent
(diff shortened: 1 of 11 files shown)


Xandaros:
Do you return the entire prompt, or just the last word?
Say I type in `/connect asd` and possible completions are `/connect asdfgh` and `/connect asdjkl` - should the function return `["asdfgh", "asdjkl"]` or `["/connect asdfgh", "/connect asdjkl"]`?
The docs mention word, but the example is only one word long, so I'm not sure.

Also, there are roughly two philosophies with completion. I personally prefer if it only completes unambiguous input (so you have to type in `/connect asdf` to complete to `/connect asdfgh`). If the input is ambiguous, it would print possible options.
I think this is better, especially if there are a lot of options - but this behaviour does not seem possible with the current implementation.
Maybe this could be added later as a setting?

LiquidityC:
That's right. The existing implementation works on just single word completion. I was also thinking about this during implementation. I'm planning on solving this. Either through two arguments to the callback, eg. Word and full line. Alternatively adding a method to retrieve the current prompt buffer through a secondary function. I'll decide once I know what's easier. I'm guessing that will be option one. The script module has very little knowledge of the outside world within the application.

LiquidityC:
Actually. My mistake. The completion callback receives the entire promptline. Hitting tab on the prompt `/connect some` will send `/connect some` as the input argument to the completion callback. So that already works the way we want it to. I need to update the documentation to explain this though.

I'll send an other commit to this PR when I get some time later.

In regards to ambiguity. Yes we could extend that further down the line it won't be critical or require modifications in this PR. It should be possible to introduce a setting to alter this behavior.

Xandaros:
Ok, good. You had me worried there :D
Not having the context of the entire line would be... suboptimal.

But yeah, the completion behaviour I personally want won't be possible with this, but it's a very good start.

Xandaros (approved):

Issue #377, opened 2021-08-23
Add Support for Submodules When Pulling Down Plugins


hjozwiak:
Describe the feature you would like added. Please be as detailed as possible.
It might be a neat idea to have support for git submodules when pulling down plugins. The main impitus for implementing this would be if a plugin ships assets as a separate git repository, such as the sound effects for a particular MUD.

PR #378, opened 2021-08-23
recursive clones impl

hjozwiak:
- Add in a helper function for helping to update submodules in a plugin's repository.
- Update the handler for add.
- Update the add_plugin alias to account for the new functionality.
- Update the updating function to take into account the fact that we can support submodules now.
- Skip over submodules that haven't been initialized when updating.
#- Update help.
If you have any suggestions for improvements, I am all ears.
#377

### DIFF
--- a/resources/help/plugin.md
+++ b/resources/help/plugin.md
@@ -6,7 +6,7 @@ of lua scripts between users. It's all based off of git.
 The following macros exist to help manually adding and loading plugins.
 
 - `/plugins`                    List installed plugins
-- `/add_plugin <url|path> -r|--recursive`      Install a plugin through a git url or file path
+- `/add_plugin <url|path>`      Install a plugin through a git url or file path
 - `/remove_plugin <name>`       Uninstall a plugin
 - `/update_plugin <name>`       Update a plugin
 - `/load_plugin <name>`         Load a plugin
--- a/resources/help/plugin.md
+++ b/resources/help/plugin.md
@@ -6,7 +6,7 @@ of lua scripts between users. It's all based off of git.
 The following macros exist to help manually adding and loading plugins.
 
 - `/plugins`                    List installed plugins
-- `/add_plugin <url|path>`      Install a plugin through a git url or file path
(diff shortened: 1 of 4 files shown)


LiquidityC (commented):
This looks good. Just one question. Is there any reason to not always do a recursive clone? I don't really see a use case where a plugin contains gitmodules that shouldn't be included with the plugin. I'm assuming that providing the recursive flag for a plugin without gitmodules has no effect.

If you agree you could still leave the lua changes as is. Just drop the --recursive flag and default the recursive call to true.

hjozwiak:
I made it opt in by default to ensure that the user doesn't get a submodule they didn't want. Why would I pull down a sound submodule on a VPS, for example?

LiquidityC:
I see that we're thinking of different scopes of plugins here. My idea of a plugin is some code that adds an isolated feature (eg. MCMP, an XP counter for Stickmud etc.). You're thinking in a broader scope where a plugin for a certain mud adds a plethora of various functionalities, am I right? Neither viewpoint is wrong.

From my viewpoint the answer to your last question would be, why would you install a sound plugin on a VPS in the first place, never mind the submodule.

In any case I'd argue that downloading a plugin that adds a plethora of different sub modules and you'd like to exclude some of them then you'd have to do that manually. As opposed to downloading a plugin and manually enabling sub modules you want included.

For a decent plugin I'm assuming the amount of users wanting to opt out of a sub module will be fewer then the amount wanting to opt in. Either way the '--recursive' flag won't exclusively solve these users problem. They will either need to do a recursive install and then manually remove sub modules or do a default install and manually include sub modules.

I just think that defaulting to recursive would be less intrusive and one less command flag. It also ensures that when you're manually installing plugins using the predefined macros then there's only one way to do it.

hjozwiak:
That is a good point, and for end users it would probably be easiest to just recursively clone the repository and strip out things they don't want to have.

LiquidityC (approved):
Nice work

PR #479, opened 2021-11-25
help: fix stale lua_gmcp.md send_gmcp reference.

cpu:
Since d938f144ac32936027165170d86de6f6a28420a6 the `gmcp.send_gmcp` method has been replaced by `gmcp.send`. One stale reference to `send_gmcp` persisted in example documentation and is fixed in this commit.


LiquidityC:
Thanks 🎉

PR #554, opened 2022-03-10
Surfacing 'clear_tags' function

beldrew:
Exposes the 'clear_tags' function to allow for its use in scripting.

### DIFF
--- a/resources/help/mud.md
+++ b/resources/help/mud.md
@@ -126,3 +126,6 @@ mud.add_tag("MSDP")
 
 The primary use for this function is to indicate what telnet protocols are
 active for the current mud.
+
+***mud.clear_tags()***
+Removes all tags from the topbar of Blightmud except for the hostname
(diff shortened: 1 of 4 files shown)


LiquidityC:
Why close?

beldrew:
A long boring story of me getting confused. 😛 I'll re-open it as it was the easiest change ever.  The function already existed just had to make it 'visible' to lua.

resources/help/mud.md
```diff
@@ -126,3 +126,6 @@ mud.add_tag("MSDP")
 
 The primary use for this function is to indicate what telnet protocols are
 active for the current mud.
+
```
LiquidityC, line 129: We need a '##' between these two specs to conform to the "format".

LiquidityC (changes requested):
All the logic looks solid. Just one minor formatting thing to fix.

resources/help/mud.md
beldrew: Oops, missed that.  Fixed

LiquidityC (approved):

LiquidityC:
Great. It just needs a `cargo fmt` now and all should go through smooth.

PR #557, opened 2022-03-10
deps: update `tts` 0.20.4 -> 0.21.0

cpu:
This commit updates the optional `tts` crate dependency used for builds with the text-to-speech feature enabled from [0.20.4](https://docs.rs/crate/tts/0.20.4) to [0.21.0](https://docs.rs/crate/tts/0.21.0).

Most notably this `tts` update pulls in an updated `speech-dispatcher`/`speech-dispatcher-sys` version that addresses a build-time incompatibility with `speechd` 0.11.1+. See [this upstream issue](https://gitlab.com/ndarilek/speech-dispatcher-rs/-/issues/3) for more information.

The net result of this update is fixing builds of Blightmud w/ text-to-speech enabled on systems that have `speechd` 0.11.1+ installed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -250,9 +250,9 @@ checksum = "a4a45a46ab1f2412e53d3a0ade76ffad2025804294569aae387231a0cd6e0899"
 
 [[package]]
 name = "bytemuck"
-version = "1.7.3"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "439989e6b8c38d1b6570a384ef1e49c8848128f5a97f3914baef02920842712f"
+checksum = "0e851ca7c24871e7336801608a4797d7376545b6928a10d32d75685687141ead"
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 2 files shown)


cpu:
I think the build failures are caused by CI using `speechd` earlier than 0.11.1. There's a "0.10" Cargo feature that I think needs to be set for tts/speech-dispatcher to build under that env. I'll take a crack at fixing shortly.

LiquidityC:
I'm going to close this because dependabot superseeded it. Looking at github CI and compilation issues I'm just going to disable `--all-features` builds for ubuntu until further notice. `libspeechd-dev` will be updated in Ubuntu 22.04 so we'll just wait until that makes it into github CI under the `ubuntu-latest` flag.

The dependabot feature is running CI as I'm writing. Ubuntu builds have already passed.

cpu:
@LiquidityC sounds good to me. Apologies I didn't end up figuring out a way to have the tts build play nicely with the two speechd versions. Thanks!

LiquidityC:
No problem. There wasn't really any obvious solution because of the default flag on tts

Issue #555, opened 2022-03-10
/dc, /disconnect sometimes stop working.


robg204:
**Describe the bug**
/dc and /disconnect often stop working for me.  Happens on both a wsl2/ubuntu x86 build, and an arm64 build on macosx. Same scripts on each host, no conflicting aliases or triggers, however seems to reproduce after longer connection times.  8-12 hours, tends to not happen under 4 hours connected.

output of /aliases just to show there are no conflicts:
  > /aliases
  [**]   42 : ^/cask$              +enabled
  [**]   40 : ^/goto \d+$          +enabled
  [**]   41 : ^/reset$             +enabled

**What terminal are you using?**
Windows Terminal, Alacritty

**What mud were you playing?**
3scapes.org 3200

**Screenshots**
Didn't screenshot, but /dc and /disconnect just do nothing.  I have not tried calling mud.disconnect() directly but will try that next time I experience this issue.

robg204:
![image](https://user-images.githubusercontent.com/49252009/157597948-11b1072d-c32c-4fa3-9b61-61a209c7fd2c.png)

LiquidityC:
Super odd. I'll see if I can figure it out based on this information.

robg204:
Connection time, it turns out, is a bit of a red herring -- but extensive use of script.reset() does seem relevant.   I validated that mud.disconnect() remains intact as a function pointer when it stops working and dug deeper.  looking at mud.rs, I noticed disconnect was passing a connection_id into event.rs, and event.rs has this:

    if self.session.connection_id() == id && self.session.connected() {

I noticed that sef.session.disconnect() is called without such intensive checking elsewhere (ie: connect), and updated my local copy to

    if self.session.connected() {

and dropped all references to passing the connection ID into the disconnect event, and everything appears fixed.  I can submit a PR with all the required changes, but I'm not sure if the passing of this connection id was in preparation for something else.

The alternative would be to figure out and understand why the connection_id is getting mangled, as I clearly when the nuclear approach of not caring about the value at all.

LiquidityC:
If I'm not completely mis-remembering. I think the connection ID is vital in the case of a reconnect `/rc`. So with the changes you made it shouldn't work to `/rc` a mud. It will just disconnect and be done with it.

robg204:
[...]

PR #561, opened 2022-03-18
remove connection_id, and it's references

robg204:
Solves #555 -- builds cleanly, /rc, /connect (both unconnected and while already connected) function as expected.

Unsure if this was being tracked for a future reason, or a historical one, however the connection id from the registry was able to get out of sync from that stored in rust, preventing /dc from working.  Would have to use /rc or /connect which both indiscriminately call session.disconnect directly without checking the ID as a conditional.  Removing the conditional check from /dc cascades into the variable not being used anywhere else, so this PR removes all references to all for a clean compile with no warnings for unused vars or dead code fns.


robg204:
In my silly haste, did not consider the the tests for cargo test, just cargo build --release.  Will update accordingly and add a commit.

LiquidityC (approved):

LiquidityC:
Nice work!

PR #573, opened 2022-04-24
Support capitalized key value in bindings

x1a0:
While migrating my bindings from TF to Blight, I noticed there is not way to bind `alt-shift-h` for example. Because `alt-H` would be normalized to `alt-h` during binding.

With this change, it's possible to bind keys like `alt-H`, which is basically `alt-shift-h`.

However I am not sure how would one bind `ctrl-alt-h` for example. I checked `Termion` and it seems not supported?

### DIFF
--- a/resources/help/bindings.md
+++ b/resources/help/bindings.md
@@ -5,7 +5,9 @@ pressed. This also include re-binding keys to a setup you are more comfortable
 with rather then the default.
 
 - `Ctrl-<char>` ex. `Ctrl-a, Ctrl-b` but not `Ctrl-PgUp`, there is no distinction for capitalization
-- `Alt-<char>` ex. `Alt-a, Alt-b` but not `Alt-PgUp`, there is no distinction for capitalization
+- `Alt-<char>` ex. `Alt-a, Alt-b` but not `Alt-PgUp`, there is distinction for capitalization.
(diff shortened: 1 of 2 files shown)


x1a0:
The failed test also fails in upstream.

LiquidityC:
I know about the tests. Some of them should be flagged 'flaky' because they are inconsistent on CI.
In regards to `ctrl-alt`, there is no easy way to bind those combinations as termion doesn't support it.

Also, this capitalization binding will only work for `alt-` bindings. Bindings using `ctrl-H` will not be respected and instead trigger the lowercase binding (`ctrl-h`, if one exists). It was a while since I looked at it but I think this is due to termions behavior also.

Your PR is still good and useful. I'd just like to see some details in the helpfile `bindings.md` about the capital situation and how it only works with `alt` bindings.

x1a0:
Thanks - I've updated the help doc.
Also I never thought it'd not work for `ctrl-{}`. After trying it and yep - it does not work. So I limit it to `alt-` only now. This means one can still define `Ctrl-H` and that'd be stored as `ctrl-h`. Otherwise it's stored as `ctrl-H` and can never be matched.

LiquidityC (approved):

PR #644, opened 2022-08-16
Consider erroring input listeners to have matched

### LINKED ISSUE #643: Input sent to MUD if alias errors
**Describe the bug**
Blightmud sends the input to the MUD if an alias matches and errors.

**To Reproduce**
Connect to a MUD and type `/lua error("test")`. The data will be sent to the MUD.

You can also try it without connecting to a MUD and it will tell you that you are not in a session.

Xandaros:
If an input listener errors, it should be considered to have matched the line.

Otherwise, any error in the input listener will cause the input to be sent to the MUD. Even if the input listener shouldn't actually match the line, it is probably best to just not send data to the MUD when an error occurred.

Fixes #643

### DIFF
--- a/src/lua/lua_script.rs
+++ b/src/lua/lua_script.rs
@@ -249,7 +249,7 @@ impl LuaScript {
     pub fn on_mud_input(&self, line: &mut Line) {
         if !line.flags.bypass_script {
             let mut lline = LuaLine::from(line.clone());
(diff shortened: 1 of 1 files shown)


LiquidityC:
Nice. This should solve the bug. If we want to get fancy we could introduce a "error" flag on a line and produce some fancy user directed error reporting. But this will do good for now.

Issue #685, opened 2022-10-27
MTTS default should include screen reader if BM started in reader mode


Xandaros:
Currently, Blightmud always sends 271 as the MTTS (in the ttype module)

If Blightmud is started in reader mode, the default should include 64 to indicate that the client is using a screen reader. The final default MTTS value for reader mode should then be 335.

It might be worth extending the ttype module with APIs to (de-)select individual parts of the MTTS value and calculate the MTTS value when it is sent, rather than storing it as a number.

LiquidityC:
We can definitely make this more modular. We'll also need to make the "mode" available to lua. 

Details about ttype here: https://tintin.mudhalla.net/protocols/mtts/

LiquidityC:
Turns out the existing options to modify MTTS are broken and never worked. The way I'm choosing to fix them will be a breaking change. My thought is that if nobody reported a bug before now we might get away with not doing a major release despite a breaking change.

But perhaps that's just conservative thinking and we should just do a major release even though very little has been added?

cpu:
> But perhaps that's just conservative thinking and we should just do a major release even though very little has been added?

Version numbers are free :-) I'd vote for frequent semver motivated releases even if it means not every major release is feature packed.

PR #686, opened 2022-10-28
Improves TTYPE module and negotiation

LiquidityC:
Fixes: #685

### DIFF
--- a/resources/lua/ttype.lua
+++ b/resources/lua/ttype.lua
@@ -1,16 +1,51 @@
 core.enable_protocol(24)
+
 local program, _ = blight.version()
 local term = os.getenv("TERM") or "xterm-256color"
-local mtts = "MTTS 271"
 
 local index = 1
 local mod = {}
+local auto_reader_mode = true
 
-local NEGOTIATION_STACK = {
-    program,
-    term,
-    mtts,
-}
+TTYPE_OPT_MTTS_ANSI =   0x001 -- Ansi support
+TTYPE_OPT_MTTS_VT100 =  0x002 -- VT100 support
+TTYPE_OPT_MTTS_UTF8 =   0x004 -- UTF-8 Support
+TTYPE_OPT_MTTS_256C =   0x008 -- 256 color support
+TTYPE_OPT_MTTS_MTRA =   0x010 -- Mouse tracking support
+TTYPE_OPT_MTTS_OSCC =   0x020 -- OSC color palette support (true color)
+TTYPE_OPT_MTTS_READ =   0x040 -- Client using screen reader
+TTYPE_OPT_MTTS_PROX =   0x080 -- This is a proxy connection
+TTYPE_OPT_MTTS_TRUC =   0x100 -- True color support
+TTYPE_OPT_MTTS_MNES =   0x200 -- Mud New Env Standard enabled
+TTYPE_OPT_MTTS_MSLP =   0x400 -- Mud Server Link Protocol enabled
+
+-- Build the default MTTS value
+local mtts = 0x0
+mtts = mtts | TTYPE_OPT_MTTS_VT100
+mtts = mtts | TTYPE_OPT_MTTS_ANSI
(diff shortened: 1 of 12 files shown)


resources/lua/ttype.lua
```diff
@@ -1,16 +1,51 @@
...
+TTYPE_OPT_MTTS_VT100 =  0x002 -- VT100 support
+TTYPE_OPT_MTTS_UTF8 =   0x004 -- UTF-8 Support
+TTYPE_OPT_MTTS_256C =   0x008 -- 256 color support
+TTYPE_OPT_MTTS_MTRA =   0x010 -- Mouse tracking support
+TTYPE_OPT_MTTS_OSCC =   0x020 -- OSC color palette support (true color)
+TTYPE_OPT_MTTS_READ =   0x040 -- Client using screen reader
+TTYPE_OPT_MTTS_PROX =   0x080 -- This is a proxy connection
+TTYPE_OPT_MTTS_TRUC =   0x100 -- True color support
+TTYPE_OPT_MTTS_MNES =   0x200 -- Mud New Env Standard enabled
+TTYPE_OPT_MTTS_MSLP =   0x400 -- Mud Server Link Protocol enabled
```
Xandaros, line 20: I'm honestly not a huge fan of this. The length of these constants is silly and I don't think they should be global.

Personally, I would strip the `TTYPE_OPT` prefix, and make these a member of the module.

I also think some of them need renaming. `PROX` and `READ` should be `PROXY` and `READER` or maybe even `SCREEN_READER`

resources/lua/ttype.lua
```diff
@@ -1,16 +1,51 @@
...
+
+local function init()
+    index = 1
+    if auto_reader_mode then
+        if tts.is_enabled() or blight.is_reader_mode() then
+            mtts = mtts | TTYPE_OPT_MTTS_READ
+        else
+            mtts = mtts & ~TTYPE_OPT_MTTS_READ
+        end
+    end
```
Xandaros, line 42: I wonder if maybe `auto_reader_mode` should be disabled if the screen reader option is manually set. The help does mention that the value gets determined automatically if the auto-detection is on, but I think it is a bit weird that you can set the option and not have it work.

Without reading the help file, it is certainly surprising.

Maybe instead of taking a bool as the value for `auto_reader_mode` in `auto_detect_reader_mode`, it should just be a function that always sets it to true, and disabling it is done by giving it a proper value? The option would essentially have three different states: on, off, auto

Xandaros (changes requested):

resources/lua/ttype.lua
LiquidityC: I work with C and low level kernel code now. Can you tell by my naming habits?

I was thinking that people might want to use the variables in scripts. But let's just make them local instead.

resources/lua/ttype.lua
LiquidityC: I'll put some thought into it.

resources/lua/ttype.lua
Xandaros: With the current API, users do need access to these. (For `add_option` and `rem_option`)

Making these local and breaking `add_option` and `rem_option` apart into a function for each option is probably the best solution. What I actually mean, though, was making these constants part of the module rather than making them global or local. (`mod.MTTS_PROXY = 0x080`)

LiquidityC: I'll move them to part of the module. Of course. My lua has gotten rusty...

resources/lua/ttype.lua
LiquidityC: I'll just make the auto_reader_mode variable toggled through `add_option` and `rem_option`. If either of these touch the screen_reader bit we will disable the auto-detection.

resources/lua/ttype.lua
LiquidityC: As for "one function per setting". It feels like more work then reward.

PR #712, opened 2022-12-28
feat: RFC 1073 (NAWS), on_dimensions_change callbacks.

cpu:
This PR adds support for Telnet "Negotiate About Window Size" (NAWS) negotiation support. See [RFC 1073](https://www.rfc-editor.org/rfc/rfc1073.html) for details.

This protocol extension allows Blightmud to communicate terminal size information to compatible MUD servers. This in turn allows servers to better size output for clients, particularly when rendering rich terminal interfaces. For Blightmud, we take care to subtract the input area and status area heights from the height transmitted to the server in NAWS so that only the drawable area is reported.

To support this feature the `blight` Lua module is updated to allow registering callbacks to be invoked whenever terminal dimensions change (`blight.on_dimensions_change`). Registered callbacks get invoked with the new terminal width/height as arguments and the NAWS module uses such a callback to update the server on the client's terminal size whenever it changes (iff NAWS has been negotiated).

### DIFF
--- a/resources/lua/naws.lua
+++ b/resources/lua/naws.lua
@@ -1,0 +1,43 @@
+-- See https://www.rfc-editor.org/rfc/rfc1073.html
+local NAWS_PROTOCOL = 31
+local naws_enabled = false
+
+-- Return table with the network byte order (e.g. big endian) encoding of the
+-- width and height given.
+local function network_dimensions(width, height)
+    bytes = {}
+    local append_byte = function(c)
+        table.insert(bytes, string.byte(c))
+    end
+    string.pack(">i2", width):gsub(".", append_byte)
+    string.pack(">i2", height):gsub(".", append_byte)
+    return bytes
+end
+
+local function send_dimensions(width, height)
+    -- We must adjust the height to just the writable area, subtracting
+    -- the size by 2 for the input/prompt area, and by the size of the status
+    -- area.
+    height = height - 2 - blight.status_height()
+    core.subneg_send(NAWS_PROTOCOL, network_dimensions(width, height))
+end
+
+-- Advertise NAWS support.
+core.enable_protocol(NAWS_PROTOCOL)
+
+-- If NAWS is negotiated update our enabled status and send the current
+-- window dimensions.
+core.on_protocol_enabled(function (proto)
+    if proto == NAWS_PROTOCOL then
+        mud.add_tag("NAWS")
+        naws_enabled = true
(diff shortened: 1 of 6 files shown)


cpu:
In 151c439 I tided up the width/height encoding using `string.pack`, but this assumes Blightmud only supports Lua 5.3+. If this isn't true it'd be better to revert that commit.

resources/lua/naws.lua
```diff
@@ -0,0 +1,39 @@
...
+-- return table with the network byte order (e.g. big endian) encoding of the
+-- width and height given.
+local function network_dimensions(width, height)
+    bytes = {}
+    local append_byte = function(c)
+      table.insert(bytes, string.byte(c))
+    end
+    string.pack(">i2", width):gsub(".", append_byte)
+    string.pack(">i2", height):gsub(".", append_byte)
+    return bytes
```
LiquidityC, line 14: Is the big endian manipulation needed? I was under the impression that this is handled in the lower network layers. However if you've tested and it works then it works.

On another note, does NAWS expect the entire terminal height or the height of the "output area". In blightmud this will be `height - 2 (for input area and prompt area) - status_area_height`.

LiquidityC (changes requested):

resources/lua/naws.lua
cpu: > Is the big endian manipulation needed? I was under the impression that this is handled in the lower network layers. However if you've tested and it works then it works.

I'll do some experimentation to see if I can remove it outright. I can confirm I did test the data transmitted with this version matches what I'd expect to see server-side for a given width/height client-side :+1: 

> On another note, does NAWS expect the entire terminal height or the height of the "output area"

Hmmm, great point. I think it's most useful if we send the height of the output area since the server wants to know the boundaries of output it can draw and that won't include the input/prompt/status areas. I'll adjust.

cpu: > I'll do some experimentation to see if I can remove it outright

It might be my own Lua inexperience but I can't find a way to create the right byte table encoding of the two ints to transmit without packing them using the stdlib string method or manually with bitwise operations.

>  I'll adjust.

I've updated the terminal height reported via NAWS to account for the input/status areas. PTAL: https://github.com/Blightmud/Blightmud/pull/712/commits/5fdd2f76b8b93f6467e50c1bc7f52b7d3e84154b

LiquidityC (approved):

PR #729, opened 2023-01-29
feat: add Lua APIs for prompt input masking.

### LINKED ISSUE #723: Create input prompt highlighting capabilities
## Description

To facilitate a spellchecker there needs to be a possibility of colorizing the input prompt content.

This should be solved by introducing a **InputPromptMask** (name isn't set in stone). This should be a struct containing ANSI escapes and indexes. Before the prompt input is rendered in the UI we should merge the prompt content with the **InputPromptMask** to create colorized output.

This way typed text and color codes don't mix which should be easier to handle. 

## Steps

- Introduce the **InputPromptMask** (mask) (choose a better name if you can think of one?)
- Merge input prompt text with the **mask** before rendering to screen
- Create lua api allowing scripts to create these **masks**
- Write tests and stuff

cpu:
### Description

This commit adds three new Lua APIs: `prompt_mask.set`, `prompt_mask.get` and `prompt_mask.clear`.

The `prompt_mask.set` function accepts prompt data, and a table comprised of integer offsets to string content to insert into the prompt data at those locations pre-rendering. Mask content is not sent through to the MUD, or added to history data. Input wrapping is handled on the rendered prompt data as if the ANSI control characters were invisible. Reader mode avoids displaying any control characters in the prompt input buffer.

The function returns a bool indicating whether the mask was set. If false, the mask may have been invalid (out of range indexes, etc), or the prompt data may have already changed. Multiple calls to `set` will combine masks, with colliding keys having their values overwritten by the most recent `set` table.

When the input prompt buffer is changed we clear the prompt mask, ensuring that the two remain in sync. When a new prompt mask is set the prompt data is freshly rendered using the overlaid mask content.

The `prompt_mask.get` function returns the prompt mask table (if any). This is the combined view of all prior calls to `prompt_mask.set` and can be used to retrieve mask data.

The `prompt_mask.clear` function removes the prompt mask (if any) and redraws the input prompt data without masking.

The primary use-case for these APIs are to support scripts adding ANSI colour codes at predefined offsets to "highlight" words (e.g. based on a spellchecking result).

See `/help prompt_mask` for more information.

Resolves #723

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ signal-hook = "0.3.14"
 mlua =  { version = "0.8.7", features = ["lua54", "send", "vendored"] }
 regex = "1.7.1"
 strip-ansi-escapes = "0.1.1"
+vte = "0.11.0"
 timer = "0.2.0"
 flate2 = "1.0.25"
 mdcat = { version = "1.0.0", default-features = false }
--- a/resources/help/prompt_mask.md
+++ b/resources/help/prompt_mask.md
@@ -1,0 +1,52 @@
+# Prompt Mask
+
+This module offers methods to mask (decorate) data that has
+been typed on the prompt line but not yet sent to the server.
+
+See also `/help prompt`.
+
+##
+
+***prompt_mask.set(data, table) -> bool***
+Set the prompt mask table to be associated with the input data. Returns true
+if the mask is valid, and the prompt data hasn't changed. Returns false if
+the mask is not valid, or the prompt data has changed and no longer matches
+the data argument.
+
+Multiple calls to `prompt_mask.set` will merge the tables, with colliding 
+keys having their value replaced by the value from the last call.
+
+- `data`    Prompt data to mask. Must match current prompt input data.
+- `table`   A Lua table of integer index keys and mask content values.
+            Each index must be a valid character index within the bounds
+            of data (1-indexed) and must fall at a character boundary.
+
+
+```lua
+-- Example: Highlight any lines that start with a dangerous command.
+prompt.add_prompt_listener(function()
+    local danger = "/quit"
+    local data = prompt.get()
+    if #data < #danger or string.sub(data, 1, #danger) ~= danger then
+        return
+    end
+    local danger_mask = {[1] = BG_RED, [#data+1] = C_RESET};
+    local res = prompt_mask.set(data, danger_mask);
+    blight.output(string.format("masked: %s",res))
+end)
+```
+
+##
+
+***prompt_mask.clear()***
+Clear the current prompt mask table (if any).
+
+##
+
+***prompt_mask.get() -> table***
+Return the current prompt mask table (if any).
+
+- `table`   A Lua table of integer index keys and mask content values
+            previously set with calls to `prompt_mask.set`.
+            Each index key will be a character index within the bounds
+            of the current prompt input data (1-indexed).
--- a/src/model/prompt_mask.rs
+++ b/src/model/prompt_mask.rs
@@ -1,0 +1,147 @@
+use mlua::{Integer as LuaInt, Lua, Result as LuaResult, String as LuaString, Table as LuaTable};
+use std::collections::BTreeMap;
+use std::ops::{AddAssign, Deref};
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct PromptMask {
+    mask: BTreeMap<i32, String>,
+}
+
+impl PromptMask {
+    pub fn new() -> Self {
+        PromptMask {
+            mask: BTreeMap::new(),
+        }
+    }
+
+    pub fn clear(&mut self) {
+        self.mask.clear()
+    }
+
+    pub fn mask_buffer(&self, buf: &[char]) -> String {
+        let mut masked_buf = buf.to_owned();
+        let mut offset = 0;
+        for (idx, mask) in self.iter() {
+            let adjusted_idx = offset + *idx as usize;
+            masked_buf.splice(adjusted_idx..adjusted_idx, mask.chars());
+            offset += mask.len();
+        }
+
+        masked_buf.iter().collect()
+    }
+
+    pub fn to_table<'a>(&'a self, ctx: &'a Lua) -> LuaResult<LuaTable> {
+        let table = ctx.create_table()?;
+        for (idx, mask) in self.iter() {
+            let adjusted_idx = *idx + 1;
+            table.set(adjusted_idx, (*mask).clone())?;
+        }
+        Ok(table)
+    }
+}
+
+impl Deref for PromptMask {
+    type Target = BTreeMap<i32, String>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.mask
+    }
+}
+
+impl AddAssign for PromptMask {
+    fn add_assign(&mut self, rhs: Self) {
+        self.mask.extend(rhs.mask)
+    }
+}
+
+impl From<BTreeMap<i32, String>> for PromptMask {
+    fn from(mask: BTreeMap<i32, String>) -> Self {
+        PromptMask { mask }
+    }
+}
+
+impl From<LuaTable<'_>> for PromptMask {
+    fn from(mask_table: LuaTable) -> Self {
+        let mut mask = BTreeMap::new();
+        for pair in mask_table.pairs::<LuaInt, LuaString>() {
+            let (offset, marker) = pair.unwrap();
+            // Lua is 1-indexed, we handle that here as part of the conversion from LuaTable.
+            // so that the Rust code can use natural 0-indexing.
+            let adjusted_offset = (offset as i32) - 1;
+            mask.insert(adjusted_offset as i32, marker.to_str().unwrap().to_string());
+        }
+        PromptMask { mask }
+    }
+}
+
+#[cfg(test)]
+mod test_prompt_mask {
+    use crate::model::PromptMask;
+    use mlua::{Lua, Table as LuaTable};
+    use std::collections::BTreeMap;
+
+    #[test]
+    fn test_from_luatable() {
+        let lua = Lua::new();
+        let simple_mask: LuaTable = lua
+            .load(
+                r#"
+    {
+        [5] = "!",
+        [1] = "*",
+    }
+"#,
+            )
+            .eval()
+            .unwrap();
+        let mut expected = BTreeMap::new();
+        // We expect the 1-indexed Lua table indices to have been translated to 0-indexing.
+        expected.insert(0, "*".to_string());
+        expected.insert(4, "!".to_string());
+
+        let lua_mask = PromptMask::from(simple_mask);
+        let expected_mask = PromptMask::from(expected);
+        assert_eq!(lua_mask, expected_mask)
+    }
+
+    #[test]
+    fn test_add_assign() {
+        let mut mask_map_a = BTreeMap::new();
+        mask_map_a.insert(10, "*".to_string());
+        mask_map_a.insert(15, "#".to_string());
+        mask_map_a.insert(20, "!".to_string());
+        let mut mask_a = PromptMask::from(mask_map_a);
+
+        let mut mask_map_b = BTreeMap::new();
+        mask_map_b.insert(1, "@".to_string());
+        mask_map_b.insert(15, "%".to_string());
+        mask_map_b.insert(25, "&".to_string());
+        let mask_b = PromptMask::from(mask_map_b);
+
+        let mut expected_map = BTreeMap::new();
+        expected_map.insert(1, "@".to_string());
+        expected_map.insert(10, "*".to_string());
+        expected_map.insert(15, "%".to_string()); // NB: overwritten by map_b.
+        expected_map.insert(20, "!".to_string());
+        expected_map.insert(25, "&".to_string());
+        let expected = PromptMask::from(expected_map);
+
+        mask_a += mask_b;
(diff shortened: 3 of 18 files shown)


src/model/prompt_mask.rs
```diff
@@ -0,0 +1,55 @@
...
+        let mut masked_buf = buf.to_owned();
+        let mut offset = 0;
+        for (idx, mask) in self.mask.iter() {
+            // NB: idx subtracted by one to account for Lua one-indexing.
+            let adjusted_idx = offset + (idx - 1) as usize;
+
+            // TODO(XXX): This is a HACK! Our mask/buf are falling out of sync somehow...
+            if adjusted_idx > masked_buf.len() {
+                return masked_buf.iter().collect::<String>();
+            }
```
cpu, line 30: I mention this in the WIP PR's TODO list. I've noticed if I enter some data with a prompt mask, then send the data, then reload the data from history, then press "down" to clear the input prompt, something weird happens:

* Pressing the down arrow ends up setting the prompt content to a new empty buffer.
* We clear the prompt mask when that happens
* But! Somehow `self.mask.iter` on L23 returns the old mask content, even though the mask has been cleared.
* This results in the `adjusted_idx` being outside the bounds of the `masked_buf` (which is empty), and a panic ensues.

I added this bounds check here as a temporary hack and it "solves" the problem. Maybe someone more familiar with the project can spot my error?

LiquidityC:
At first glance this looks great. I'm glad the idea with a prompt mask had the intended low impact and avoiding having to re-invent the command buffer to make this coloring work.
I'll take a peek and see if I can't see what's causing that buffer thing also. As for prompt wrapping. I don't remember how we draw it. But just skip all masks that are "before" the index where we render the prompt input from. That might be trickier then it sounds. As I said, don't really recall how it's implemented :smile:

cpu:
quick update: found the cause of the desync and implemented a fix locally. I'm looking at the prompt wrapping part next.

> skip all masks that are "before" the index where we render the prompt input from. 

makes sense. I think I see a way to make this work but it will require pushing the prompt mask down into the UI code that draws the prompt since that's where the wrapping happens.

On a related note: I'm thinking `ReaderScreen`'s impl of the `UserInterface` trait's `print_prompt_input` fn should ignore the masking entirely. Does that make sense to you or should I try to support that UI type as well?

LiquidityC:
> quick update: found the cause of the desync and implemented a fix locally. I'm looking at the prompt wrapping part next.
> 
> > skip all masks that are "before" the index where we render the prompt input from.
> 
> makes sense. I think I see a way to make this work but it will require pushing the prompt mask down into the UI code that draws the prompt since that's where the wrapping happens.

Yes. I hate that... If it starts to get messy I think we should take a step back and thing.
Perhaps we can create an iterator that only steps through printable chars perhaps? I know the prompt wrapping relies on the `chars()` iterators a lot.

> On a related note: I'm thinking `ReaderScreen`'s impl of the `UserInterface` trait's `print_prompt_input` fn should ignore the masking entirely. Does that make sense to you or should I try to support that UI type as well?

Yes. I see no situation where someone is using reader mode without actually relying on a screen reader. This is not the first case where we omit "colors" in `ReaderScreen`

cpu:
> Perhaps we can create an iterator that only steps through printable chars?

@LiquidityC This idea worked nicely, thanks! PTAL at the implementation. I'll work on adding docs + tests next. I think it will be ready to go after that.

<details>
<summary>I've been testing with this simple lua script to highlight words in the prompt area that are an even number of characters:</summary>

```lua
local red_on = "\x1b[31m";
local ansi_off = "\x1b[0m";

prompt.add_prompt_listener(function (buf)
    local words = {}
    local word_start = 1
    for i = 1, #buf do
        local c = buf:sub(i, i)
        local stop = c:find("[%s%p]")

        if stop ~= nil then
            local word = buf:sub(word_start, i - 1)
            words[word_start] = word
            word_start = i + 1
        elseif i == #buf then
            local word = buf:sub(word_start, i)
            words[word_start] = word
        end
    end

    local mask = {}
    for idx, word in pairs(words) do
        local even = #word % 2 == 0
        local off_idx = idx + #word
        if even then
            mask[idx] = red_on
            mask[off_idx] = ansi_off
        end
    end

    prompt_mask.set(buf, mask)
end)

```
</details>

<details>
<summary>Here's a video of the feature in action, showcasing history + prompt wrapping all working nicely. No more hacky workarounds:</summary>

[asciicast](https://asciinema.org/a/v9QDne59s6AJUYcQgg8G9KFPs)

</details>

Cargo.toml
```diff
@@ -25,6 +25,7 @@ signal-hook = "0.3.14"
 mlua =  { version = "0.8.7", features = ["lua54", "send", "vendored"] }
 regex = "1.7.1"
 strip-ansi-escapes = "0.1.1"
+vte = "0.11.0"
```
cpu, line 28: quick note: I added this `vte` dependency to be able to parse out terminal control codes in `printable_chars.rs`. However, it is [already a transitive dependency via the `strip-ansi-escapes` crate](https://github.com/luser/strip-ansi-escapes/blob/a0a306ab6c5a76f269fea8659fdd736f1d2d56b9/Cargo.toml#L13-L14) so there's no increase in the net size of deps.

LiquidityC: No problem

LiquidityC:
> > Perhaps we can create an iterator that only steps through printable chars?
> 
> @LiquidityC This idea worked nicely, thanks! PTAL at the implementation. I'll work on adding docs + tests next. I think it will be ready to go after that.

I'm really glad it worked. The impact on the original code is very small now which is always comforting. I'll take a closer look tomorrow. Nice work.

LiquidityC:
With some tweaking once this is in place it would be possible to write a plugin that highlights alias matches in the prompt input. So that you know if you are sending an alias before you're actually sending it. Perhaps not valuable. But it's cool.

cpu:
:wave: I force pushed some revisions. Some highlights:

* I added docs, under `/help prompt_mask`.
* I added support for calling `prompt_mask.set` multiple times. The mask tables are added together and colliding keys are replaced by the most recently provided value. 
* I added validation of the mask table. Providing a mask table with out of bound indexes or an index that doesn't fall at a clean character boundary will return false.
* There are unit tests now. A couple are a bit rough, so I might try to polish them up while this is under review but the core functionality is all tested.

<details>
<summary>Here's an updated test.lua that uses two masks:</summary>

```lua
-- convert buf to a table of words, indexed by their start location in buf.
function word_table(buf)
    local words = {}
    local word_start = 1
    for i = 1, #buf do
        local c = buf:sub(i, i)
        local stop = c:find("[%s%p]")

        if stop ~= nil then
            local word = buf:sub(word_start, i - 1)
            words[word_start] = word
            word_start = i + 1
        elseif i == #buf then
            local word = buf:sub(word_start, i)
            words[word_start] = word
        end
    end

    return words
end

-- set a prompt mask for buf where all even length words are highlighted red.
function even_word_mask(buf)
    local words = word_table(buf)

    local mask = {}
    for idx, word in pairs(words) do
        local even = #word % 2 == 0
        if even then
            local hi_on = BG_RED .. C_BWHITE
            local off_idx = idx + #word
            mask[idx] = hi_on
            mask[off_idx] = C_RESET
        end
    end

    prompt_mask.set(buf, mask)
end

-- set a prompt mask for buf where special words are highlighted green.
function special_word_mask(buf)
    local words = word_table(buf)

    local mask = {}
    for idx, word in pairs(words) do
        if word == "blightmud" then
            local hi_on = BG_GREEN .. C_BWHITE
            local off_idx = idx + #word
            mask[idx] = hi_on
            mask[off_idx] = C_RESET
        end
    end

    prompt_mask.set(buf, mask)
end

-- set up two prompt listeners for setting prompt masks.
prompt.add_prompt_listener(even_word_mask)
prompt.add_prompt_listener(special_word_mask)
```

</details>

<details>
<summary>Here's an updated demo video:</summary>

[asciicast](https://asciinema.org/a/fvy2y6rnybVZI68IohGM7iURA)
</details>

I'm going to leave this a draft PR until I've had a chance to do that last bit of polish but it's ready for review whenever you have time. Thanks!

LiquidityC:
I think this looks great. Nice work. I just have one request for added functionality. After that you can mark it as ready for review and we'll get this bad boy merged.

- Introduce a `prompt_mask.clear()` in the lua API
- Introduce a `prompt_mask.exists()` in the lua API

These two functions should make it at least possible for two separate "masking-plugins" to play nicer and not disrupt each other too much. If it becomes a major issue with colliding masks we can extend this to have "instanced" masks or something. But I don't see any point in making it that general yet.

cpu:
> Introduce a prompt_mask.clear() in the lua API

Done.

> Introduce a prompt_mask.exists() in the lua API

This one turns out to be a bit tricky. I can stash a `bool` for "mask exists" from the `set` API impl w/ `set_named_registry_value`, but when the mask is invalidated by user input there's no convenient way I can spot from the `command.rs` context to update that value to `false`, meaning the result will be stale if a prompt mask is set and then new input arrives and the prompt mask is cleared.

I also put some thought into whether we could make a nicer `prompt_mask.get()` API that would return the table. This ends up being tricky because we can't stash the user provided table in the `set()` call (e.g. like how `prompt.set` supports `prompt.get`) because it wouldn't be the full "combined" result. When we have the combined mask result in hand we're in the command context where there's no easy access to the Lua backend to update a named registry value.

I might be missing an easy solution. If not, I think we could get by without `exists/get`. WDYT?

src/ui/command.rs
```diff
@@ -415,13 +439,13 @@ pub fn spawn_input_thread(session: Session) -> thread::JoinHandle<()> {
...
+                                        buffer.get_pos(),
+                                    ))
+                                    .unwrap();
                             }
-                            writer
-                                .send(Event::UserInputBuffer(
-                                    buffer.get_buffer(),
-                                    buffer.get_pos(),
-                                ))
-                                .unwrap();
```
cpu, line 424: Quick note on this change: while testing I noticed that key binding events (e.g. F1, etc) would send a `UserInputBuffer` event even though none of the events `check_command_binds` processes could provide new input. This meant that if you set a prompt mask from a key binding it would immediately be invalidated when the unchanged prompt input is sent as a `UserInputBuffer` event.

Moving where the event was emitted so it is conditional on `check_command_binds` returning `false` fixed my test script so that I could press "f4" and have the highlights appear, where previously pressing "f4" would briefly flash the mask before it was invalidated and cleared:


```lua
blight.bind("f3", function ()
    blight.output("clearing mask");
    prompt_mask.clear()
end)
blight.bind("f4", function ()
  local data = prompt.get()
  if #data == 0 then
    blight.output("no prompt data")
    return
  end
  local danger_mask = {[1] = BG_RED, [#data+1] = C_RESET};
  local res = prompt_mask.set(data, danger_mask);
  blight.output(string.format("masked: %s",res))
end)
```

cpu:
> I might be missing an easy solution.

Oh, I think the solution is to add a method to `LuaScript` that does the needful with the Lua `state`. The command context has access to the session and the LuaScript inside. I'll give that a whirl after lunch. Sorry for the noise.

LiquidityC:
If there's no simple solution then skip it. Clear is the important one. The other is just QoL

cpu:
> If there's no simple solution then skip it. Clear is the important one. The other is just QoL

I just pushed a revision that adds `prompt_mask.get`. It wasn't too bad :+1:

LiquidityC:
> > Test & Build / Tests (ubuntu-latest, nightly) (pull_request) Failing after 4m
> 
> ```
> error: test failed, to rerun pass `--test communication_tests`
> 
> Caused by:
>   process didn't exit successfully: `/home/runner/work/Blightmud/Blightmud/target/debug/deps/communication_tests-345bac71090d49cc` (signal: 11, SIGSEGV: invalid memory reference)
> Error: Process completed with exit code 101.
> ```
> 
> I ran these tests locally ~2000 times and didn't catch this failure. Maybe this is similar flakyness to #726? thinking
> 
> _Edit: Further flake evidence: [it happened](https://github.com/Blightmud/Blightmud/actions/runs/4138419422/jobs/7154807856) against tip of dev testing a docs only PR_

Yes. I haven't been able to figure out what's causing it. But it's rare enough now that I can't really be bothered either. If you have any best guess to what the problem is then please share :smile:

resources/help/prompt_mask.md
```diff
@@ -0,0 +1,50 @@
...
+    local res = prompt_mask.set(data, danger_mask);
+    blight.output(string.format("masked: %s",res))
+end)
+```
+
+##
+
+***prompt_mask.clear()***
+Clear the current prompt mask table (if any).
+
```
LiquidityC, line 43: Nitpick: According to the "common pattern" there's usually a `##` separator between doc sections.

src/ui/command.rs
LiquidityC: Looks like you fixed a potential future bug in my eyes.

LiquidityC (changes requested):
All the code looks great. I just have one comment on one of the help files and it's a massiv nit-pick. Other then that everything looks great.
An added bonus is that there are almost no red sections (removals) in the diff which makes the chance for breaking existing functionality almost zero.

I think we cooked up a good solution here and you wrote good code. Thanks!

cpu:
> I think we cooked up a good solution here and you wrote good code.

Thanks for the help iterating on the design :-)  Just wanted to note that pushing the docs fix un-did the auto-merge.

LiquidityC (approved):

PR #735, opened 2023-02-09
docs: add simple debug guide to CONTRIBUTING.md

cpu:
### Description

Since Blightmud is a rich terminal application it's a little bit tricky to debug it interactively while still keeping the client usable.

This commit adds some simple instructions for using `gdbserver` to debug Blightmud in one terminal while running it in another.

I use a similar setup under the CLion IDE but think it's better to give the generic GDB advice in the docs. Users can adapt to their IDE of choice with a little bit of legwork.


LiquidityC (approved):

cpu:
> the :Termdebug functionality in vim which essentially does this setup inside vim

TIL :sparkles:

Issue #737, opened 2023-02-11
feat request: Lua API for prompt cursor position


cpu:
### Description

It would be useful if the `prompt` (or `blight`?) module could be augmented with a function to retrieve the current prompt input cursor position offset, and a function to set the prompt input cursor position to a specific offset. 

It's already possible for Lua code to _move_ the cursor by using `blight.ui` to emit relevant events like `"step_left"` and `""step_word_left"`, but I don't see any way for Lua code to find the current position in the input prompt. I would also like more fine-grain control, e.g. being able to set the cursor position to a specific numeric offset in the prompt data vs having to navigate with UI events.

### Example use-cases

* I'd like to augment the [Blightspell](https://github.com/cpu/blightspell) plugin to support keybindings that would allow moving the cursor between misspelled words. I might be able to rig something up with UI events for stepping between words, but since I already know the offset of the words I want to navigate to it would be much easier to directly set the cursor position.

* I'd like to augment the [Blightspell](https://github.com/cpu/blightspell) plugin to support replacing a misspelled word under the input cursor with a suggested alternative. This requires knowing if the user has positioned the input cursor over a misspelled word, meaning I need to know the cursor position.

### Planning

Happy to try to implement this feature myself, just wanted to file an issue so we can discuss the shape of the solution before I spend too much time writing it. It may be worth including this issue in [the overarching spellchecking project](https://github.com/orgs/Blightmud/projects/1).

LiquidityC:
These all sound like reasonable and useful API methods to provide. I don't see any issue with them.

LiquidityC:
I can take a look at this today actually.

PR #740, opened 2023-02-13
Adds lua functions to fetch and manipulate prompt cursor pos

LiquidityC:
Fixes: #737

### DIFF
--- a/resources/help/prompt.md
+++ b/resources/help/prompt.md
@@ -14,6 +14,25 @@ Returns the line currently typed into the prompt
 
 ***prompt.set(input)***
 Sets the line typed in the prompt. Replacing any current data.
+This set the cursor position to the end of the new prompt.
+
+##
+
+***prompt.get_cursor_pos() -> number***
+Gets the cursor position in the input prompt
+
+- Returns the current index of the cursor starting at 0
+
+##
+
+***prompt.set_cursor_pos(pos)***
+Sets the cursor position in the input prompt.
+
+- `pos` A positive number
+
+Attempting to set the position to a value larger then the length of the current
+prompt input will move the cursor to the end of the prompt input.
+Attempting to set the position to a negative value will triggger an error.
 
 ##
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -332,6 +332,18 @@ For more info: https://github.com/LiquidityC/Blightmud/issues/173"#;
                         .unwrap();
                 }
             }
+            Event::SetPromptCursorPos(pos) => {
+                if let Ok(mut buffer) = session.command_buffer.lock() {
+                    buffer.set_pos(pos);
+                    session
+                        .main_writer
+                        .send(Event::UserInputBuffer(
+                            buffer.get_buffer(),
+                            buffer.get_pos(),
+                        ))
+                        .unwrap();
+                }
+            }
             Event::ServerSend(_)
             | Event::ServerInput(_)
             | Event::Connect(_)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -335,12 +335,12 @@ For more info: https://github.com/LiquidityC/Blightmud/issues/173"#;
             Event::SetPromptCursorPos(pos) => {
                 if let Ok(mut buffer) = session.command_buffer.lock() {
                     buffer.set_pos(pos);
+                    if let Ok(mut script) = session.lua_script.lock() {
+                        script.set_prompt_content(buffer.get_buffer(), buffer.get_pos());
+                    }
                     session
                         .main_writer
-                        .send(Event::UserInputBuffer(
-                            buffer.get_buffer(),
-                            buffer.get_pos(),
-                        ))
+                        .send(Event::UserInputCursor(buffer.get_pos()))
                         .unwrap();
                 }
             }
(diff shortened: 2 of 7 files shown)


resources/help/prompt.md
```diff
@@ -14,6 +14,25 @@ Returns the line currently typed into the prompt
...
 ***prompt.set(input)***
 Sets the line typed in the prompt. Replacing any current data.
+This set the cursor position to the end of the new prompt.
+
+##
+
+***prompt.get_cursor_pos() -> number***
+Gets the cursor position in the input prompt
+
+- Returns the current index of the cursor starting at 0
```
cpu, line 24: I think it might be nice to have this API return a 1-indexed value since most users will immediately do that conversion themselves to work with the prompt data as a Lua string. In the prompt mask API I converted to/from the Lua style indexing in the UserData impl that sits at the edge of the Rust/Lua code.

src/lib.rs
```diff
@@ -333,6 +333,18 @@ For more info: https://github.com/LiquidityC/Blightmud/issues/173"#;
...
             }
+            Event::SetPromptCursorPos(pos) => {
+                if let Ok(mut buffer) = session.command_buffer.lock() {
+                    buffer.set_pos(pos);
+                    session
+                        .main_writer
+                        .send(Event::UserInputBuffer(
+                            buffer.get_buffer(),
+                            buffer.get_pos(),
+                        ))
```
cpu, line 344: I added `Event::UserInputCursor` in anticipation of this API as part of https://github.com/Blightmud/Blightmud/pull/738. I think it'd be better to use that here since the buffer contents haven't changed, we just want a redraw of the prompt at the new cursor pos.

```suggestion
                        .send(Event::UserInputCursor(buffer.get_pos()))
```

resources/help/prompt.md
```diff
@@ -14,6 +14,25 @@ Returns the line currently typed into the prompt
...
+
+##
+
+***prompt.set_cursor_pos(pos)***
```
cpu, line 28: If you agree with my suggestion to use Lua indexing we'll also want to account for that w/ `set_cursor_pos(pos)` for symmetry.

src/lua/lua_script.rs
```diff
@@ -279,10 +280,12 @@ impl LuaScript {
         result.ok()
     }
 
-    pub fn set_prompt_content(&mut self, content: String) {
+    pub fn set_prompt_content(&mut self, content: String, pos: usize) {
         self.exec_lua(&mut || -> LuaResult<()> {
             self.state
                 .set_named_registry_value(PROMPT_CONTENT, content.clone())?;
+            self.state
+                .set_named_registry_value(PROMPT_CURSOR_INDEX, pos)?;
```
cpu, line 288: We also need to update the `PROMPT_CURSOR_INDEX` registry value after processing command binds since those directly manipulate the buffer's pos. Right now it seems like the value quickly gets out of sync with the cursor position.

Here's a script I used to demonstrate the problem:
```lua
blight.bind("f5", function ()
    blight.output("moving cursor");
    prompt.set_cursor_pos(4)
end)
blight.bind("f6", function ()
    local pos = prompt.get_cursor_pos()
    blight.output("cursor pos: ".. tostring(pos));
end)
```

1. If you type some input and press `f6` the position at the end of the buffer is correctly reported :heavy_check_mark: 
2. If you press `f5` the cursor is moved to position 4 and the prompt redrawn :heavy_check_mark: 
3. If you press `f6` again, the position is still reported as the end of the buffer :x: 

I think you'll need to add a `luascript.rs` function to set the registry value, and then call that from `command.rs` in the `termion::event::Event::Key` match arm that I was mucking with in https://github.com/Blightmud/Blightmud/pull/738. I think we want to update the prompt cursor value in the place that emits `UserInputCursor` and the place that emits `UserInputBuffer`.

cpu (changes requested):

src/lua/lua_script.rs
cpu: I had a rough WIP branch with logic that accounts for the above if its helpful: https://github.com/Blightmud/Blightmud/commit/1354844e84054ea367cc61f5e2740c90c81570d2.

resources/help/prompt.md
cpu: Mmm, I see your point :+1:

src/lua/lua_script.rs
LiquidityC: Nice spot. I'll dig into it

resources/help/prompt.md
LiquidityC: I have no strong opinions on this. Let's go with the Lua format here.

src/lib.rs
LiquidityC: Done

resources/help/prompt.md
LiquidityC: Done

cpu (approved):
Revisions look good to me modulo the docs fix :rocket: Thanks!

resources/help/prompt.md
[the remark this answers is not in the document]
cpu: Oh wait, this is the thing we talked about yesterday. It returns 0 for an empty buffer.

PR #773, opened 2023-03-29
Adds a macos-11 build to the rust workflow

LiquidityC:


mirageglobe:
Looks good!

LiquidityC:
Yes. Om going to move this to the release build before I merge then I think we'll do a patch release. It's just slow progress. Been swamped at work lately.

PR #775, opened 2023-04-01
net: replace native-tls with rustls.

cpu:
## Description

This branch replaces the dependency Blightmud uses for TLS MUD connections. The [`native-tls`](https://github.com/sfackler/rust-native-tls) crate is replaced with [`rustls`](https://github.com/rustls/rustls). Rustls has the advantage of being a pure Rust implementation of TLS 1.2 and TLS 1.3, requiring no unsafe/native code except in the underlying *ring* dependency that provides cryptographic primitives.

### Trusted root certificates

Since we're no longer delegating to the OS specific system TLS implementation via `native-tls` we must provide our own root certificate store. For this purpose I opted to use the [`webpki-roots`](https://github.com/rustls/webpki-roots) crate. It provides a reasonable set of default root certificates that closely follows Mozilla's curated root program, matching Firefox's trusted CAs. 

### Unit Tests

I also added test coverage for the TLS connection handling. Doing this requires setting up a realistic server certificate and private key, which in turn requires a CA certificate and private key to use to sign the leaf server certificate. I opted to use [`minica`](https://github.com/jsha/minica) to generate these test resources because it's approximately 1000x easier than alternatives. This is only required for developers that want to update/recreate the test certs. I left the instructions required in a README alongside the test certs/key material.

### Real world testing/compatibility

I've tested this branch against `dunemud.net:6788` and `stickmud.com:7670`. Rustls may be stricter about standards compliance than rustls-native was on some platforms, so it would be prudent to test this code as best as we can with other MUDs. 

Along similar lines, Rustls doesn't implement the insecure legacy SSLv3 or TLS 1.0/1.0 protocols, only TLS 1.2 and TLS 1.3. I don't expect this to cause much trouble in the MUD space since so few MUDs have even bothered to set up TLS - I doubt those that have done so would have done it using outdated and insecure defaults. 

If we want to be extremely conservative about breaking any existing users we could look at supporting _both_ native-tls and rustls for a period. I experimented with that based on a feature flag [in this earlier WIP branch](https://github.com/Blightmud/Blightmud/commit/c4a3c6a143647153acec27b109243553527f269b).

### Future work

Blightmud still requires OpenSSL after this branch lands due to the [`git2-rs`](https://github.com/rust-lang/git2-rs) and [`curl`](https://docs.rs/curl/latest/curl/) dependencies, but it is no longer used when communicating to MUD servers. I think there is a path forward to replace ~both~ the curl dependency and will pursue this in follow up work. Replacing git2-rs with something like gitoxide is likely blocked on submodule support.


cpu:
> Blightmud still requires OpenSSL after this branch lands due to the [git2-rs](https://github.com/rust-lang/git2-rs) and [curl](https://docs.rs/curl/latest/curl/) dependencies, but it is no longer used when communicating to MUD servers. I think there is a path forward to replace both the curl dependencies and will pursue this in follow up work. 

I have a PR built on top of this branch that will replace `curl` with `reqwest + rustls` for a pure Rust alternative for version checking: https://github.com/cpu/Blightmud/pull/1 I'll open that PR on this repo once the `rustls` work is in-tree.

I looked into replacing `git-rs` and unfortunately it doesn't look like Gitoxide can clone submodules so we're not able to replace `git-rs` for plugin management yet. It looks like Gitoxide is undergoing rapid development so hopefully that will change in the future and we can revisit.

LiquidityC:
This is awesome work. I trust that you've tested what you can and I think the next step is just to let this run on dev for a while and see if any bugs shake out of the tree.
If you rebase it I can merge it.

src/net/tcp_stream.rs
```diff
@@ -81,7 +82,12 @@ pub fn spawn_connect_thread(
                 tls,
                 verify_cert,
             } = connection;
-            if !session.connect(&host, port, tls, verify_cert) {
+            let tls_validation = if verify_cert {
```
LiquidityC, line 85: Can we perhaps do a `impl From<bool> for CertificatValidation` here?

cpu: :+1: Implemented in my recent force push.

cpu:
> I think the next step is just to let this run on dev for a while and see if any bugs shake out of the tree. If you rebase it I can merge it.

SGTM :-) Rebased.

PR #782, opened 2023-04-09
net: rewrite version check using reqwest.

cpu:
Previously the `net/check_version.rs` implementation used to warn users at runtime about newly released Blightmud versions relied on Rust bindings for the native `curl` library. This was the only place in the codebase using that dependency.

Instead of relying on a crate based on native code, this commit replaces the implementation with [`reqwest`](https://docs.rs/reqwest/latest/reqwest/) using the `rustls` TLS backend we implemented for MUD TLS. Using reqwest w/ rustls provides end-to-end HTTP(S) requests in pure Rust without needing `libcurl` or `libssl`.

Since we're performing version checks in a background thread I've implemented this check using the blocking reqwest client.

Along the way I made some refactoring cleanups:

* `diff_version` isn't required, we can use `&str.cmp` directly.
* `Fetcher`'s error `Result` was never meaningfully used, so switch to `Option`.
* Rather than returning `Vec<u8>` and manually deserializing fields, the `Fetcher` trait is switched to return a struct with a derived `Deserialize` and the impl uses reqwest's `Response.json` to unmarshal the HTTP response into it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,7 +238,6 @@ version = "5.2.0"
 dependencies = [
  "anyhow",
  "chrono",
- "curl",
  "env_logger",
  "flate2",
  "getopts",
@@ -256,10 +255,11 @@ dependencies = [
  "notify-debouncer-mini",
  "pulldown-cmark",
  "regex",
+ "reqwest",
  "rodio",
  "ron",
  "rs-complete",
- "rustls",
+ "rustls 0.21.0",
  "rustls-pemfile",
  "serde",
  "serde_json",
@@ -271,7 +271,7 @@ dependencies = [
  "timer",
  "tts",
  "vte 0.11.0",
- "webpki-roots",
+ "webpki-roots 0.23.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


cpu:
@LiquidityC Interesting news: I spent some time this afternoon trying to dig into the random SIGSEGV errors we see in CI sometimes and miraculously this branch fixes them entirely. I was even able to uncomment the `#[ignore]`'s on the other communication related tests and see 0 failures, even running the tests in bulk. 

That was enough to give me a hint that the problem must be related to the version checking code and curl-rust. To verify I set up some modified CI for tip of `dev` that ran the tests many times per run and was able to fairly reliably reproduce SIGSEGV errors. I then modified tip of `dev` to comment out the call to `check_latest_version` from `lib.rs`'s `run` func. That was enough to have CI pass all tests without SIGSEGV, even in bulk, confirming the problem must be related to the curl-rust based version checking. Based on some [upstream issues like this one](https://github.com/alexcrichton/curl-rust/issues/333) I suspect it has to do with a combination of native library initialization order and the libcurl/openssl versions installed. I was never able to reproduce these crashes locally with very modern libcurl, and in CI I only ever see the Ubuntu based tasks fail with SIGSEGV, never MacOS.

It might be possible to further isolate the bug but since I had already re-written the problematic code in this PR I vote we leave it a bit of a mystery, merge this PR, restore the ignored tests, and move on with our lives :-) I've run tests with this branch over and over and have seen zero crashes total.

LiquidityC:
Nice

Issue #744, opened 2023-02-16
Re-think the CTRL-C exit cycle


LiquidityC:
The ctrl-c cycle (needing to submit it twice) is sensitive to other events arriving between the first and second ctrl-c (sigint) signal. The new event SetPromptMask was not included in the whitelist of events that can be received and would interupt the ctrl-c sequence.

This system should probably be re-worked to use some form of timer or at least highlight that the sequence has been interrupted. It's otherwise quite easy for a plugin to unknowingly interrupt this cycle.

LiquidityC:
A timer functionality here would be immune to interrupting events but I haven't really dug into how tricky it would be to achieve.

Another option could just be to indicate to the user that the cycle has been interrupted. This would make it clear to scripters/plugin devs that something is getting in the way. However if you're not using ctrl-c to exit then you won't notice. And it's also awkward since the lua functionality might be a required one and there is no direct way of excluding the event from interrupting.

Thoughts?

PR #789, opened 2023-04-13
robust ctrl-c quit confirmation.

cpu:
Previously the quit confirmation that made sure users didn't exit Blightmud unexpectedly was based on looking at events received between the first ctrl-c press and the confirmation. This required keeping an allow list of ignored events (`UserInputBuffer`, `TimedEvent`, `TimerTick`, `SetPromptMask`, and `ClearPromptMask`) that shouldn't clear the confirmation flag when received. In practice keeping that list up to date as new event types are added is problematic (e.g. https://github.com/Blightmud/Blightmud/pull/743) and so a more robust alternative is needed.

This commit replaces the quit confirmation with a new system where the confirmation is cleared by a dedicated event (`QuitConfirmTimeout`) sent by a separate thread after waking up from a 5s sleep. The first ctrl-c sets the pending quit flag and spawns the thread. A second ctrl-c being received before the timeout thread wakes up will quit Blightmud. If the thread wakes up before a second confirmation ctrl-c is sent then it writes a `QuitConfirmTimeout` event that will result in the pending quit flag being cleared. If the user presses ctrl-c at this point it will start the cycle again.

Resolves https://github.com/Blightmud/Blightmud/issues/744

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -11,9 +11,11 @@ use crate::{
 };
 use libtelnet_rs::{bytes::Bytes, events::TelnetEvents};
 use log::debug;
+use std::thread::JoinHandle;
 use std::{
     error::Error,
     sync::mpsc::{channel, Receiver, Sender},
+    thread, time,
 };
 
 #[derive(Debug, PartialEq, Eq, Clone)]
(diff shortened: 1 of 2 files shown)


LiquidityC (approved):

LiquidityC:
This looks good. The failing test is because of a concurrency problem that for some reason has flown under the radar for quite some time. I'm patcing it right now. I suggest you just hold off for a bit and rebase when it's in.

LiquidityC:
#790 is awaiting review and should fix the test issue you're having.

PR #790, opened 2023-04-13
Fixes concurrency issue in integration tests

LiquidityC:
The `setup()` function used to prepare a local blightmud instance when running integration tests would always load the script `tests/timer_test.lua`. This script sends some data over the wire which could in some cases interupt other telnet negotiation tests. This should no longer be the case now.


cpu:
Nice catch :-)

PR #801, opened 2023-04-28
feature: make spellcheck dependency optional.

cpu:
This commit adds a `spellcheck` feature flag that is default-enabled, and makes the `hunspell-rs` and `hunspell-sys` dependencies optional based on that feature flag.

This will allow most users to get spellchecking by default, but for the select few having trouble building `hunspell-sys` on ARM64, they can use `--no-default-features` to avoid the problem.

It might also be worth adding CI coverage for build/test with `--no-default-features` but CI time is getting quite long with the recent additions so for now I've left it out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,15 @@ commands that might fit your system:
 - Ubuntu    `apt install libclang-dev libspeechd-dev speech-dispatcher speech-dispatcher-espeak espeak`
 - Arch      `pacman -S speech-dispatcher espeak`
 
+### Compile without spellchecking
+
+Some users may encounter issues building the spellcheck feature on MacOS ARM64 (M1/M2). To
+build Blightmud without the spellcheck feature, use `--no-default-features`. E.g.:
+
+- Install rust
+- Run `cargo build --no-default-features` or `cargo build --no-default-features --features text-to-speech`
+- Run `cargo run --no-default-features` or `cargo run --no-default-features --features text-to-speech` to run
+
 ## Installation
 
 - **Ubuntu/Debian**      : Deb packages can be found on the releases page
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -84,6 +84,15 @@ commands that might fit your system:
 - Ubuntu    `apt install libclang-dev libspeechd-dev speech-dispatcher speech-dispatcher-espeak espeak`
 - Arch      `pacman -S speech-dispatcher espeak`
 
+### Compile without spellchecking
+
+Some users may encounter issues building the spellcheck feature on ARM64. To
```
LiquidityC, line 89: Be more specific here. It's not ARM64 it's Mac ARM64. This chip has caused me a lot of headaches so let's not "not point fingers". :smile:

src/lua/lua_script.rs
```diff
@@ -10,7 +10,9 @@ use super::{
 use crate::lua::fs::Fs;
 use crate::lua::prompt::Prompt;
 use crate::lua::prompt_mask::PromptMask;
+#[cfg(feature = "spellcheck")]
 use crate::lua::spellcheck;
+#[cfg(feature = "spellcheck")]
 use crate::lua::spellcheck::Spellchecker;
```
LiquidityC, line 16: Group these includes so that we only need one feature flag:
`use crate::lua::spellcheck::*;`
Or the squiggly bracket version. I forget the pattern.

LiquidityC:
I had a few comments besides that  I'm open to taking this in because it does in fact seem like a simple thing to make a feature out of. Features can get really awkward even in rust.

I also prefer not having to deal with Mac M1 issues.

PR #807, opened 2023-05-06
feat: add echo_input setting.

cpu:
Previously all `ServerInput` events unconditionally called `screen.print_send` to echo input sent by a user to the screen. In the split screen mode this results in the input appearing with a `> ` prefix and special colouring. That's a sensible default but I would be happy to rely on the input buffer history for remembering what I sent and save some output screen space by not echoing all my input.

This commit adds a `echo_input` setting that can be changed to disable screen input echo behaviour on an opt-in basis.

The setting defaults to true, and can be changed at runtime with immediate effect. When set to false input is still logged but will no longer appear on the screen with a `> ` prefix. Setting it to true will restore the echo effect.

### DIFF
--- a/resources/help/scrolling.md
+++ b/resources/help/scrolling.md
@@ -3,6 +3,10 @@
 You can scroll the output history of your game using keys (default
 `PgUp`/`PgDn`) or mouse (if you enabled the setting).
 
+You can scroll through commands you sent to the game using
+`Ctrl PgUp`/`Ctrl PgDown` if `/set echo_input on` is enabled (default). 
+See `/help search` for more information.
+
 If your current output area is longer then 20 lines then blightmud will split
 the window into two, the upper will show you the output history that you are
 scrolling and the lower will show you the live output from your mud.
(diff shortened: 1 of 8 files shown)


LiquidityC:
This one looks fine but it will break `ctrl - pgup/pgdn` behavior (see https://github.com/Blightmud/Blightmud/blob/dev/resources/help/search.md).
So make a note of that in the help text please.
Since I think it's a feature that often goes overlooked perhaps you could add the info about `ctrl - pgup/pgdn` in `scrolling.md` within this PR since you're laready fiddling with it?

LiquidityC (changes requested):
See my comment in the PR.

cpu:
> This one looks fine but it will break ctrl - pgup/pgdn behavior (see [search.md](https://github.com/Blightmud/Blightmud/blob/dev/resources/help/search.md)).
> So make a note of that in the help text please.

Good catch, thank you. I've updated the `/help search` text. I also patched the Lua `search` module so `search.find_next_input` and `search.find_last_input` will print an error message if `echo_input` is disabled.

> Since I think it's a feature that often goes overlooked perhaps you could add the info about ctrl - pgup/pgdn in scrolling.md within this PR since you're laready fiddling with it?

Done.

LiquidityC (approved):
Nice work. Thanks

PR #821, opened 2023-05-30
Replaces mdcat dependency with pulldown-cmark-mdcat

LiquidityC:


cpu:
Is the rationale for the change that `pulldown-cmark-mdcat` is the lower-level alternative to `mdcat` and so a better fit for Blight or did something else motivate it?

LiquidityC:
mdcat dropped the library functionality from the their crate and pointed users to this crate for just the library functionality. The mdcat crate is just the binary now.

cpu:
>  The mdcat crate is just the binary now.

Ahhh ok :+1:

PR #926, opened 2023-10-29
Fixes for multi-byte UTF-8 data in GMCP messages

cpu:
## Description

I happened to notice that UTF-8 emoji in Telnet subnegotiation GMCP data was causing strange behaviour. For example, `Comm.Channel` events with a JSON payload where a value contained :wave: (e.g. `[0xF0 0x9F 0x91 0x8B]`) would result in receiving part of the `Comm.Channel` event as a garbled GMCP message, with only the content after the `0xF0` delivered to the Lua handler. I wasn't able to reproduce this using [Mudlet](https://www.mudlet.org/) - its GMCP support faithfully delivered the `Comm.Channel` content without issue.

Ultimately there were two bugs to be fixed, one in `libtelnet-rs` and one in Blightmud:

1. `libtelnet-rs` needs to be fixed to not improperly terminate telnet subnegotiations. Presently it's treating `0xF0` in a subnegotiation as the Telnet command `SE`  "End of subnegotiation parameters." even when it isn't preceded by `0xFF` (IAC).
2. the `gmcp.lua` `_subneg_recv` func in Blightmud was using `utf8.char(unpack(data))` to convert a table of int byte values to a utf8 string. The manner this function works does not play well with multi-byte characters like the :wave: emoji. The fix here is to use a simpler `_utf8_from` fn that can pass through the content without mangling emoji.

## Libmudtelnet

[My PR](https://github.com/envis10n/libtelnet-rs/pull/21) to fix the `libtelnet-rs` bug hasn't seen any response. I also found [an unrelated panic](https://github.com/envis10n/libtelnet-rs/pull/22), and a third bug related to IAC unescaping. Since it seems like the original maintainer doesn't have a lot of time, the crate needs touch ups, and Blightmud is the only dependent listed in crates.io, I forked it into the Blightmud org under a new name [libmudtelnet](https://github.com/blightmud/libmudtelnet). This fork has three bug fixes, was rewritten to avoid indexing that may panic, and was fuzz tested against the original implementation (with bug fixes as required) to ensure compatibility between the two.

This branch replaces Blightmud's dependency on `libtelnet-rs` with the forked `libmudtelnet`. This will also unblock merging this work immediately.

## Reproduction

The game sends the GMCP data for a message with text including an emoji:

```
Comm.Channel.Text { "text": "Paradox : 👋\n", "channel_ansi": "\u001b[1m[\u001b[0m\u001b[31mchat\u001b[0m\u001b[1m]\u001b[0m", "talker": "Paradox", "channel": "chat" }
```

With tip of `dev`, e.g. neither of the above fixes applied the GMCP data gets corrupted entirely, even the `[GMCP]` echoing fails.

```
> /lua gmcp.echo(true)
> /lua gmcp.register("Comm.Channel")
> chat :wave:
���\n", "channel_ansi": "\u001b[1m[\u001b[0m\u001b[31mchat\u001b[0m\u001b[1m]\u001b[0m", "talker": "Paradox", "channel": "chat" }[1m[chat] Paradox : 👋
```

With only the `libtelnet-rs` fix applied the GMCP message is properly parsed and echoed, but the emoji in the `text` has been mangled by `utf8.char` into `ð`:

```
> /lua gmcp.echo(true)
> /lua gmcp.register("Comm.Channel")
> chat :wave:
[GMCP]: Comm.Channel.Text { "text": "Paradox : ð\n", "channel_ansi": "\u001b[1m[\u001b[0m\u001b[31mchat\u001b[0m\u001b[1m]\u001b[0m", "talker": "Paradox", "channel": "chat" }
[chat] Paradox : 👋
```

And finally with both fixes applied we get the correct GMCP message with the correct `text`, :tada:

```
> /lua gmcp.echo(true)
> /lua gmcp.register("Comm.Channel")
> chat :wave:
[GMCP]: Comm.Channel.Text { "text": "Paradox : 👋\n", "channel_ansi": "\u001b[1m[\u001b[0m\u001b[31mchat\u001b[0m\u001b[1m]\u001b[0m", "talker": "Paradox", "channel": "chat" }
[chat] Paradox : 👋
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
 dependencies = [
  "gimli",
 ]
@@ -19,9 +19,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.3"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "86b8f9420f797f2d9e935edf629310eb938a0d839f984e25327f3c7eed22300c"
+checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
 dependencies = [
  "memchr",
 ]
@@ -65,9 +65,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.1"
+version = "0.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6cd65a4b849ace0b7f6daeebcc1a1d111282227ca745458c61dbf670e52a597"
+checksum = "2ab91ebe16eb252986481c5b62f6098f3b698a45e34b5b98200cf20dd2484a44"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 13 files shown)


LiquidityC:
Excellent work. I'll see if I can nudge the maintainer of libtelnet-rs through back-channels :smile:

cpu:
> I'll see if I can nudge the maintainer of libtelnet-rs through back-channels

Cool, thanks! I think I can figure out a unit testing strategy but probably not until the weekend :-)

cpu:
As promised I pushed a unit test. Running it without 33064d3 fails as expected:
```
thread 'lua::lua_script::lua_script_tests::test_gmcp_utf8' panicked at src/lua/lua_script.rs:1487:9:
assertion `left == right` failed
  left: "ð\u{9f}\u{91}\u{8b}"
 right: "👋"
```
Afterwards it passes:
```
test lua::lua_script::lua_script_tests::test_gmcp_utf8 ... ok
```

cpu:
@LiquidityC I forked `libtelnet-rs` into [Blightmud/libmudtelnet](https://github.com/Blightmud/libmudtelnet) so this can be merged if you're happy with it. The upstream bugfix PR doesn't seem like its moving along and Blightmud is the only consumer of that crate listed in crates.io. I'm happy to maintain it myself as part of Blightmud work so this branch is updated to switch to the fork. There are three separate bugs the fork addresses (see [CHANGELOG.md](https://github.com/blightmud/libmudtelnet/blob/main/CHANGELOG.md)) and while I've done a fair amount of work to clean up the code generally I've avoided breaking changes and tested it for compatibility with `libtelnet-rs` (minus differences produced based on the bug fixes) so it should be a no-op for users.

LiquidityC:
Let's give it some more time. The maintainer has historically been very responsive and helpful so I'd like to give him another week to respond. I'll make some pushes to get a response.

LiquidityC (approved):

PR #937, opened 2023-11-11
net: configure TCP keepalive, prefer IPv6, more debug info

cpu:
This branch updates the `open_tcp_stream` fn from `net/util.rs` to configure the opened socket with [TCP keepalive](https://en.wikipedia.org/wiki/Keepalive#TCP_keepalive) enabled.

This TCP feature is particularly important for MUDs because Telnet has no consistently enabled application layer keepalive feature. Without a keepalive mechanism long-lived and idle Telnet connections are prone to being closed by middlebox routers. When this happens Blightmud won't detect the socket was closed until the user tries to read or write from the socket. Other MUD clients [like Mudlet](https://github.com/Mudlet/Mudlet/blob/31ea3079e63735a344379e714117e4f1ad6b2f1b/src/ctelnet.cpp#L3052-L3138) configure TCP keepalive to compensate for this using a native (but typically default off) TCP feature.

In order to achieve TCP keepalive for Blightmud we take a new dependency on [`socket2`](https://docs.rs/socket2/latest/socket2/), a crate that abstracts over unsafe OS-specific mechanisms for customizing socket options. We use `socket2` to configure the opened socket to start sending TCP keepalives 30 seconds after a connection is detected as idle. Keepalives will be sent every 5 seconds, and we'll tolerate up to 5 lost keepalives before deciding the connection is closed.

We default to configuring TCP keepalive but a new "keepalive_enabled" setting can be set to 'false' by users that don't want this feature for some reason.

The address resolution logic is also changed to prefer IPv6, but fall-back to IPv4. We now interleave IPv6/IPv4 addresses to maximize the chance of finding a working address quickly, while still preferring IPv6 if available. This is a rough aproximation of the "happy eyeballs" approach used by browsers.

Lastly, to aid in future debugging new `debug!` log lines are added for which addresses are resolved, which are tried, and errors encountered trying to connect. Using the debug log level means these won't be present unless Blightmud is run with `-V` for verbose mode.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -272,6 +272,7 @@ dependencies = [
  "serde_json",
  "signal-hook",
  "simple-logging",
+ "socket2 0.5.5",
  "strip-ansi-escapes",
  "syntect",
  "termion",
@@ -1256,9 +1257,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.147"
+version = "0.2.150"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b4668fb0ea861c1df094127ac5f1da3409a82116a4ba74fca2e58ef927159bb3"
+checksum = "89d92a4743f9a61002fae18374ed11e7973f530cb3a3255fb354818118b2203c"
 
 [[package]]
 name = "libgit2-sys"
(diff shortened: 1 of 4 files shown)


LiquidityC (approved):
Nice work

PR #941, opened 2023-11-16
Adds telnet::GA to the supported list of telnet opts

LiquidityC:
For some reason GA (249) wasn't included in this list. Which caused blightmud to reply IAC DONT GA to the very few muds that actually negotiated this.


cpu (approved):
Nice catch :+1:

PR #942, opened 2023-11-16
Adds "graceful" handling of protocol disable negotation

LiquidityC:
Since the standard compliance on most mud codebases is "iffy" at best we never really introduced any handling for IAC WONT PROTO negotiation. Although compliance hasn't really changed among the various codebases it will avoid confusion if Blightmud would correctly pick up on and "disable" protocols locally when the mud requests it.

To facilitate this, functionality to listen for protocol disable requests as well as removing tags (topbar tags) have been introduced across the board.

### DIFF
--- a/src/net/telnet.rs
+++ b/src/net/telnet.rs
@@ -21,6 +21,8 @@ pub struct TelnetHandler {
     main_writer: Sender<Event>,
     output_buffer: Arc<Mutex<OutputBuffer>>,
     mode: TelnetMode,
+    will_ga: bool,
+    will_eor: bool,
 }
 
 impl TelnetHandler {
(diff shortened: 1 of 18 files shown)


cpu:
> LiquidityC requested a review from cpu 4 hours ago

ACK'ing the request. Just leaving a note that it might take me a couple days.

src/net/telnet.rs
```diff
@@ -30,11 +32,33 @@ impl TelnetHandler {
...
             output_buffer: session.output_buffer,
             mode: TelnetMode::UnterminatedPrompt,
+            will_ga: false,
+            will_eor: false,
         }
     }
-}
 
-impl TelnetHandler {
+    fn determine_telnet_mode(&mut self) {
```
cpu, line 40: small naming nit: `determine` makes me think it would be idempotent and return the new telnet mode as opposed to mutating `self` and setting the mode for `self` and `output_buffer`.

```suggestion
    fn update_telnet_mode(&mut self) {
```

cpu (approved):
Looks great, nice work :-)

Before merging could you update `resources/help/core.md` to mention `core.on_protocol_disabled` alongside the existing coverage of `core.on_protocol_enabled`?

src/net/telnet.rs
LiquidityC: Sure

LiquidityC:
Ooof, forgot about the docs. :facepalm:

PR #945, opened 2023-11-19
Move scroll default bindings to lua

LiquidityC:
The default bindings home, end, pgup, pgdn were implemented in rust. At
the same time these bindings are perfectly fine to do in lua but weren't
documented as such or implemented.

This patch moves the default scroll bindings to lua and documents them
in the helpfile. They have also been removed from the rust code since it
serves no purpose leaving them in there.

It is better to limit the "rust bindings" to core commands that aren't
bindable by users.


cpu (commented):
Nice :+1:

Issue #1125, opened 2024-10-16
How to auto reconnect on "Connection refused"?


x1a0:
There are `on_connect` and `on_disconnect`. However it seems neither is invoked when the connection attempt fails, which makes sense...

LiquidityC:
Don't see why not. It's susceptible to looping but that will be up to the user to handle.

PR #1126, opened 2024-10-16
Add `ConnectionFailed` event

x1a0:
A draft for https://github.com/Blightmud/Blightmud/issues/1125.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -36,6 +36,7 @@ pub enum Event {
     ClearTimers,
     Connect(Connection),
     Connected(u16),
+    ConnectionFailed,
     DisableProto(u8),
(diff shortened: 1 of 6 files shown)


LiquidityC (approved):
Looks good to me

LiquidityC:
It's been a while since I looked at this. But what remains is an actual connection error detection to trigger the callback. Should be findable if you start looking in mud_connection.rs. The connect function must lead to where these events originate.

LiquidityC:
Here:
https://github.com/Blightmud/Blightmud/blob/f89a96fc7ac4910c1aa608c77a1792bed887673a/src/session.rs#L65-L72

PR #1322, opened 2026-01-30
feat: add startup parameter --codec to specify MUD server codec

rascalrr:
### What's changed
Add a new startup parameter --codec that allows users to manually specify the character encoding codec for the connected MUD server.
### Usage
Users can specify the codec via startup command (e.g., --codec gbk, --codec big5) to adapt to MUD servers with different encoding standards, which resolves the text garbling issue caused by inconsistent default encoding.
### Key Notes
Supported codecs: Mainstream MUD common encodings (gbk, big5, etc.), following standard character encoding naming rules
Backward compatible: The program will use the original default codec if the --codec parameter is not specified, no impact on existing usage
### Type of change
-[x] New feature (non-breaking change which adds functionality)
-[ ] Bug fix (non-breaking change which fixes an issue)
-[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-[ ] Documentation update

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -233,6 +233,7 @@ version = "5.4.0"
 dependencies = [
  "anyhow",
  "chrono",
+ "encoding_rs",
  "env_logger",
  "flate2",
  "getopts",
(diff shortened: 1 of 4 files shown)


LiquidityC:
I assume this fixes the issue where chinese and some russian muds are not playable?

PR #1333, opened 2026-01-31
Replace network impl with mio event loop.

### LINKED ISSUE #788: Blightmud immediately disconnecting when connecting to TLS mud
@cpu yeah it's not crashing anymore 👍 but with tls enabled i now get instantly disconnected when directly launched via terminal. i'll test this later on linux also.

steps to reproduce:
```
> blightmud -n -t -c mg.mud.de:4712
```
results in:
grafik

but sometimes it works and it stays conneted. without tls enabled, it always seems to work. maybe its a timing issue.

_Originally posted by @cmetz in https://github.com/Blightmud/Blightmud/issues/621#issuecomment-1501180067_

LiquidityC:
Replaces the shared (and unsafe) split socket implementation with a single thread mio event loop (poll).

This still needs some stress testing. Completely untested right now. But if it works it will fix #788

### DIFF
--- a/src/net/event_loop.rs
+++ b/src/net/event_loop.rs
@@ -295,7 +295,7 @@ impl NetworkEventLoop {
                         if event.is_readable() {
                             if let Err(e) = self.do_read() {
                                 if e.kind() != ErrorKind::WouldBlock {
-                                    debug!("Read error, closing connection: {}", e);
+                                    error!("Read error, closing connection: {}", e);
                                     self.shutdown = true;
                                     break;
                                 }


cmetz:
Maybe we also can update mio to version 1.1.1, but so far it runs quite smoothly. i'll do some more testing. And the compression seems to work. My mud is reporting the following:

```
Telnet mode  : on
Terminal size:
  Current MUD setting: 125x26
  Info from terminal : 125x26
MXP support  : No
MSP support  : No
MCCP support : Yes
MCCP Version 2 active.
   90937 bytes compressed to 7935 bytes. (91.27% savings)
TLS encryption: Yes
        Cipher: TLS_CIPHER_AES_256_GCM
  Key exchange: TLS_KX_ECDHE_RSA
   Compression: TLS_COMP_NULL
Last lag measurement: 0.030 seconds.
Ok.
```

And also great work. :+1:

PR #1406, opened 2026-04-01
Adds tag support to lines through lua.

LiquidityC:
This is just the initial step for the new tag feature. Remaining pieces
are:

- Opt out of tag rendering (setting)
    * Important since tags indent output 2 spaces.
- Re-drawing the view with only history lines matching a given 'key'.
    * Needs a lua function to enables/disable this filter
- Setting for the default tag symbol. Lua function?

### DIFF
--- a/resources/help/bindings.md
+++ b/resources/help/bindings.md
@@ -105,4 +105,10 @@ blight.bind("up", history.previous_command)
 blight.bind("down", history.next_command)
 blight.bind("ctrl-p", history.previous_command)
 blight.bind("ctrl-n", history.next_command)
+
+-- Toggle tag rendering
+blight.bind("ctrl-t", function()
+    print("Toggling tags")
+    blight.show_tags(not blight.show_tags())
+end)
 ```
--- a/resources/help/blight.md
+++ b/resources/help/blight.md
@@ -158,8 +158,19 @@ Pass `nil` to clear the symbol filter.
 
 ##
 
+***blight.filter_tag_reverse([val]) -> bool***
+Get or set the filter reverse flag. When `false` (default), lines matching the
+filter are hidden. When `true`, lines that do *not* match the filter are hidden
+(i.e. only matching lines are shown). Pass `nil` or omit to get the current
+value without changing it.
+
+- `val`  `true` to reverse filter behavior, `false` to restore default behavior.
+- Returns the current value of the reverse flag.
+
+##
+
 ***blight.filter_tag_reset()***
-Clears all active tag filters.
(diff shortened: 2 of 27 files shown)


x1a0:
Thanks for drafting this!

When I coded it, there were 2 ideas I wanted to implement but not sure how:
- Make the tag column toggle-able. Then with some command like `/tagcol` one can turn it on/off. Notice this does not mean enable/disable tagging. The tagging function is always there as long as tag values are set for the `line`. It's a pure UI toggle - so one could hide it easily and only show it when necessary. I think this means the tag column will need to be rendered separately to the normal output area. And somehow it needs to align on the "line number" between tag column and the main output area. I am not sure how to approach that.
- This feature naturally works under the trigger system. However most triggers gag the line and print something instead. In my implementation this will silently forfeit the tag properties if they are added to the line, due to there is no output of the line any more. By reading your code I believe it'd do the same? I think it'd be great if tag could be kept somehow but I cannot see how. Maybe if the idea above is implemented this would be possible too.

LiquidityC:
Good idea with the togglable tag column. It will have to be part of the
rendered line since the whole output view is a scrolling area and those
just function on the vertical.

I think we can just redraw the output view when you toggle between tags
and no tags.

In this impl the tags live on the model/line.rs object which AFAICR is
what we store in the history buffer. I'm planning on leveraging this for
filtering and re-drawing.

Regarding GAG:ed line. Perhaps they are excluded from the history
buffer, but we can change that. That way all data stays in buffer. So
technically you could un:GAG GAG:de lines when filtering on a tag.
Not sure I'll build that into this PR. But the data will be available
for extensions later on.

Did all this make sense? A lot of text and I'm not always the best
explainer in the medium of text.

x1a0:
@LiquidityC Yes - sounds good to me. Thanks! 👍

LiquidityC:
Switched the history to store `model::Line` now. Pressing `ctrl-t` now toggles lines on and off through `blight.show_tags(true)`. Only tested this roughly. It's a lot of code.

Please test if you have time.

lua typedefs have also been updated to handle the new methods.

LiquidityC:
It's now possible to filter the output (applies on existing lines). Check the `blight` and `line` help files.

LiquidityC:
`/help tags` should give a good overview now.

LiquidityC:
Currently it's only possible to set one filter at a time. Is this to constraining? Personally I don't see a huge value in being able to set multiple... But it's one of those things that feel like they are guaranteed to expand eventually.

Since changing this might require an API change perhaps it's better to get it done now? Thoughts?

PR #1421, opened 2026-04-21
Show/hide prompt bar

### LINKED ISSUE #1420: Setting to show/hide prompt bar
The current prompt is locked in position at the bottom of the split screen.  This is a nice behavior as it certainly reduces spew, however some folks are used to the prompt being added to the scroll buffer. This behavior allows a person to see prompt changes (hp, mana, etc..) over time.

An incoming pull request, for your consideration, provides a new setting which allows users the option to toggle showing or hiding the prompt.  Hiding it, adds the prompt text to the scroll buffer.

karvanitis:
PR introducers ability for users to toggle the prompt to show and hide.  Hiding the prompt adds prompt line to the scroll back buffer.

https://github.com/Blightmud/Blightmud/issues/1420

### DIFF
--- a/resources/help/settings.md
+++ b/resources/help/settings.md
@@ -50,6 +50,7 @@ Available settings are:
 - `tts_enabled`         Enable tts (only if compiled with TTS).
 - `reader_mode`         Switches to a screen reader friendly TUI. (Does not support `status area`).
 - `hide_topbar`         Toggles the topbar.
+- `hide_prompt`         Toggles the prompt.
 - `echo_input`          Toggles whether user input is echoed on-screen with a `> ` prefix.
 - `last_command`        Toggles whether last command is persisted for easy repeat submission.
 
--- a/resources/help/settings.md
+++ b/resources/help/settings.md
@@ -50,7 +50,7 @@ Available settings are:
 - `tts_enabled`         Enable tts (only if compiled with TTS).
 - `reader_mode`         Switches to a screen reader friendly TUI. (Does not support `status area`).
 - `hide_topbar`         Toggles the topbar.
-- `hide_prompt`         Toggles the prompt.
+- `capture_prompt`      Toggles prompt capturing.
 - `echo_input`          Toggles whether user input is echoed on-screen with a `> ` prefix.
 - `last_command`        Toggles whether last command is persisted for easy repeat submission.
 
--- a/src/model/settings.rs
+++ b/src/model/settings.rs
@@ -23,10 +23,10 @@ pub const COMMAND_SEARCH: &str = "command_search";
 pub const SMART_HISTORY: &str = "smart_history";
 pub const ECHO_INPUT: &str = "echo_input";
 pub const LAST_COMMAND: &str = "last_command";
-
+pub const HIDE_PROMPT: &str = "hide_prompt";
 pub const KEEPALIVE_ENABLED: &str = "keepalive_enabled";
 
-pub const SETTINGS: [&str; 14] = [
+pub const SETTINGS: [&str; 15] = [
     LOGGING_ENABLED,
     TTS_ENABLED,
     MOUSE_ENABLED,
@@ -40,6 +40,7 @@ pub const SETTINGS: [&str; 14] = [
     SMART_HISTORY,
     ECHO_INPUT,
     LAST_COMMAND,
+    HIDE_PROMPT,
     KEEPALIVE_ENABLED,
 ];
 
@@ -78,6 +79,7 @@ impl Default for Settings {
         settings.insert(SMART_HISTORY.to_string(), false);
         settings.insert(ECHO_INPUT.to_string(), true);
         settings.insert(LAST_COMMAND.to_string(), true);
+        settings.insert(HIDE_PROMPT.to_string(), false);
         settings.insert(KEEPALIVE_ENABLED.to_string(), true);
         Self { settings }
     }
--- a/src/model/settings.rs
+++ b/src/model/settings.rs
@@ -23,7 +23,7 @@ pub const COMMAND_SEARCH: &str = "command_search";
 pub const SMART_HISTORY: &str = "smart_history";
 pub const ECHO_INPUT: &str = "echo_input";
 pub const LAST_COMMAND: &str = "last_command";
-pub const HIDE_PROMPT: &str = "hide_prompt";
+pub const CAPTURE_PROMPT: &str = "capture_prompt";
 pub const KEEPALIVE_ENABLED: &str = "keepalive_enabled";
 
 pub const SETTINGS: [&str; 15] = [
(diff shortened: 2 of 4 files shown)


src/model/settings.rs
```diff
@@ -23,10 +23,10 @@ pub const COMMAND_SEARCH: &str = "command_search";
 pub const SMART_HISTORY: &str = "smart_history";
 pub const ECHO_INPUT: &str = "echo_input";
 pub const LAST_COMMAND: &str = "last_command";
-
+pub const HIDE_PROMPT: &str = "hide_prompt";
```
LiquidityC: We probably want to call this "lock_prompt"? We aren't hiding the prompt. Just not dedicating it to a particular line. So 'lock_prompt' or 'capture_prompt' would be better. Default to true.

resources/help/settings.md
```diff
@@ -50,6 +50,7 @@ Available settings are:
 - `tts_enabled`         Enable tts (only if compiled with TTS).
 - `reader_mode`         Switches to a screen reader friendly TUI. (Does not support `status area`).
 - `hide_topbar`         Toggles the topbar.
+- `hide_prompt`         Toggles the prompt.
```
LiquidityC: 'Toggles prompt capturing'

src/ui/split_screen.rs
```diff
@@ -194,8 +195,9 @@ impl UserInterface for SplitScreen {
         if width > 0 && height > 0 {
             self.width = width;
             self.height = height;
-            self.output_line = height - self.status_area.height() - 2;
+            self.output_line = height - self.status_area.height() - if settings.get(HIDE_PROMPT)? { 1 } else { 2 };
             self.mud_prompt_line = height - self.status_area.height() - 1;
+            self.mud_prompt_enabled = !settings.get(HIDE_PROMPT)?;
             self.prompt_line = height;
```
LiquidityC, line 201: Perhaps make `mud_prompt_line: Option<u16>` instead and drop `mud_prompt_enabled`?

src/ui/split_screen.rs
```diff
@@ -266,7 +268,11 @@ impl UserInterface for SplitScreen {
     fn print_prompt(&mut self, prompt: &Line) {
         //debug!("UI: {:?}", prompt);
         self.mud_prompt = prompt.clone();
-        self.redraw_prompt();
+        if self.mud_prompt_enabled {
+            self.redraw_prompt();
+        } else {
+            self.print_output(prompt);
+        };
     }
```
LiquidityC, line 276: This approach won't work for character mode muds. If we receive content in multiple packets. Whatever line we receive last in the last packet will be displayed as a prompt and suffixed with a `\r\n`.

See my comment in the linked issue.

You could possibly move this logic into `redraw_prompt()` and print the unterminated content there...
I really can't reason about it without testing. 

If you test this out on a mud that doesn't send `IAC,GA` or `IAC,EOR` (tagged with GA/EOR in Blightmud top banner) you will see artifacts in output when a lot of content is received.

LiquidityC (changes requested):
The approach is OK and I'm fine with bringing something like this into the fold. We get the request every once in a while and I always forget to keep an example trigger of how to do it in Lua handy. The benefit with this is that we don't waste the prompt line as an empty line.

However. I assume this is only tested on a "terminated prompt" mud that sends GA or EOR. A change like this needs to function flawlessly even on character mode muds.
This is most likely why it was never tackled in core implementation before.

src/ui/split_screen.rs
karvanitis: I understand your point regrading specific muds, thus capturing the prompt overall is a really good feature. For muds which do send the end of transmission signal, however, it could be considered subjective. This can be somewhat resolved by making it a user preference with capture being the default behavior as it benefits the most people.

Personally I see this feature as shifting responsibility of interpreting what is or is not the prompt to the end user based solely on the packets generated by the mud, as intended or otherwise.  But to be fair I only play on a GA enabled mud...

src/ui/split_screen.rs
```diff
@@ -666,17 +664,25 @@ impl SplitScreen {
...
+
         write!(
             self.screen,
             "{}{}{}",
-            termion::cursor::Goto(1, self.mud_prompt_line),
+            termion::cursor::Goto(1, self.mud_prompt_line.unwrap()),
             termion::clear::CurrentLine,
             self.goto_prompt(),
         )
         .unwrap();
```
LiquidityC, line 678: Wrap this in `if let` block instead of the early return.

src/ui/split_screen.rs
LiquidityC: This needs to work nicely for both types of mud otherwise I consider it a regression and just plainly introducing this functionality using the trigger system is a better solution. Which also shifts the responsibility to the end user.

src/ui/split_screen.rs
```diff
@@ -195,9 +195,9 @@ impl UserInterface for SplitScreen {
         if width > 0 && height > 0 {
             self.width = width;
             self.height = height;
-            self.output_line = height - self.status_area.height() - if settings.get(HIDE_PROMPT)? { 1 } else { 2 };
-            self.mud_prompt_line = height - self.status_area.height() - 1;
-            self.mud_prompt_enabled = !settings.get(HIDE_PROMPT)?;
+            self.output_line = height - self.status_area.height() - if settings.get(CAPTURE_PROMPT)? { 2 } else { 1 };
+            self.mud_prompt_line = if settings.get(CAPTURE_PROMPT)? { Some(height - self.status_area.height() - 1) } else { None };
+            self.status_area_line = height - self.status_area.height();
```
LiquidityC, line 200: You can extract the setting to a local variable. Use `setting.then()` pattern for the second case where mud_prompt_line is set.

src/ui/split_screen.rs
karvanitis: From my perspective, the trigger mechanism is deficient as it continues to render the prompt bar unnecessarily leading to redundant text being rendered, the prompt is essentially rendered twice. The intent here is to alter the behavior of the client, not to merely append redundant text to the output buffer, which itself cannot be accomplished with a trigger.

Can we describe what the intended behavior is for both types of muds? I tried it on various muds that do not support GA and the behavior was as intended.

LiquidityC: You can gag the prompt. You don't need to render it twice:

```lua
trigger.add("^.*$", { prompt = true, gag = true }, function(_, line)
	mud.output(line:raw())
end)
```
You lose the one line which is the prompt since it no longer renders the prompt. This approach might still produce the artifacts described above. But in this case it's self inflicted which I can't control.

karvanitis: Thanks for sharing the trigger. I will leave it for you to consider further.

PR #1426, opened 2026-04-27
Make rodio optional

### LINKED ISSUE #1295: Make the rodio dependency optional
When using blightmud in a shell, the audio support is quite redundant. Also, some shared shell hostings might not even offer the alsa libraries as an option.

The rodio depedency could be behind a feature flag.

vegai:
- Adds an audio feature (default-on) that gates the rodio dependency, so Blightmud can be built on hosts without ALSA libs
- When the feature is off, src/audio/stub.rs provides no-op Player, SourceOptions, and handle_audio_event so lib.rs and the Lua bindings need no cfg branches
- README documents the no-audio build, and the Linux CI job now covers spellcheck-only, no-features, and default builds

Resolves #1295.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -43,9 +43,12 @@ jobs:
         with:
           toolchain: ${{ matrix.toolchain }}
           components: clippy, rustfmt
-      - name: Test
+      - name: Test feature combinations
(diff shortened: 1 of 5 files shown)


LiquidityC (approved):
Looks good. Thanks

PR #1428, opened 2026-04-28
Add `ConnectionFailed` event

x1a0:
I mentioned this in #1125 some time ago and implemented it locally.

### DIFF
--- a/resources/help/mud.md
+++ b/resources/help/mud.md
@@ -84,6 +84,19 @@ end)
 
 ##
 
+***mud.on_connect_failed(callback)***
+Registers a callback that is triggered when a connection attempt fails.
+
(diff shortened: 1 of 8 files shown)


LiquidityC (commented):
This looks good. I'd like a test and some documentation for the end user before I merge though.

LiquidityC (commented):
I just recalled one more thing. It's new so I forgot it and I apologize for not bringing it up in my first comment.

We now have a file `resources/lua/types/blightmud.d.lua` which contains lua type definitions which can be leveraged by editors or LSPs. I'd like your new function `on_connect_failed` to be included in that.

Once that's done I'll merge this.

LiquidityC (approved):

PR #1434, opened 2026-05-20
Timestamp debug logfiles

emilyaviva:
Currently, when running in verbose mode (`--verbose` or `-V`), the logfile gets written to the same place: the hardcoded path `./.run/data/logs/log.txt`. This means that if you run Blightmud a second time in verbose mode, the second log will clobber the first; furthermore, running more than one instance of Blightmud verbosely will cause them both to write to the same logfile.

This PR timestamps the logfiles when they are being created, as: `./.run/data/logs/log.<YYYYMMDD>.<HH:MM:SS>.txt`

This has the effects that you can run more than one Blightmud at a time in verbose mode, and also that you can safely run a verbose a second time, secure in the knowledge that your first logfile won't go anywhere.

This PR also updates `CONTRIBUTING.md` with a command to tail the output of just the most recently-created verbose logfile.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,10 +86,18 @@ cargo build --features=tts && \
     ./target/debug/blightmud -V
 ```
 
+Running Blightmud in verbose mode (`--verbose` or `-V`) writes a timestamped logfile at `./.run/data/logs/log.<YYYYMMDD>.<HH:MM:SS>.txt` (or, in a release build, at `$XDG_DATA_HOME/blightmud/logs/`).
+
+This command tails the output of the mosst recently created such logfile:
+
 ```bash
 # Terminal B - Blightmud debug log
-tail -F ./.run/data/logs/log.txt
+tail -F "$(ls -t ./.run/data/logs/log.*.txt | head -n1)"
 ```
+
+If you have several debug sessions running at once, target the specific file you want.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -86,11 +86,19 @@ cargo build --features=tts && \
     ./target/debug/blightmud -V
 ```
 
+Running Blightmud in verbose mode (`--verbose` or `-V`) writes a timestamped logfile at `./.run/data/logs/log.<YYYYMMDD>.<HH:MM:SS>.txt` (or, in a release build, at `$XDG_DATA_HOME/blightmud/logs/`).
```
LiquidityC, line 89: Is this true? Is the timestamp exclusive to verbose mode?

LiquidityC (commented):
Looks good, and very useful. Thanks.

Just one question. If my question is valid I'd prefer if that statement was actually true so that timestamps are restricted to verbose mode. This way debugging users get multiple logs but casual users don't fill up a folder with logs that nver get looked at.

CONTRIBUTING.md
emilyaviva: Really good catch, thanks! This is incorrect; `start_logging` doesn't actually care about verbose mode as it stands. New commit will fix this shortly…

LiquidityC (approved):

PR #1436, opened 2026-05-22
Allow customizing the content of the top bar

krendil:
- Added new lua function `blight.top_line` to set the text content of the top line.
  - Text will be integrated into the bar in the same manner as bar lines in the status area
  - Passing an empty string will result in an unbroken bar
  - Passing `nil` will reset to the default content (server hostname and tags)
  - Text that is too long will be truncated to fit within the bar

Current top bar behaviour should be unchanged if the user doesn't call this function.

### DIFF
--- a/src/ui/split_screen.rs
+++ b/src/ui/split_screen.rs
@@ -181,6 +181,7 @@ pub struct SplitScreen {
     prompt_input_pos: usize,
     show_tags: bool,
     tag_mask: TagMask,
+    top_line: Option<String>,
 }
 
 impl UserInterface for SplitScreen {
@@ -567,6 +568,11 @@ impl UserInterface for SplitScreen {
         Ok(())
     }
 
+    fn set_top_line(&mut self, line: Option<String>) -> Result<()> {
+        self.top_line = line;
+        self.redraw_top_bar()
+    }
+
     fn flush(&mut self) {
         self.screen.flush().unwrap();
     }
@@ -616,6 +622,7 @@ impl SplitScreen {
             prompt_input_pos: 0,
             show_tags: false,
             tag_mask: TagMask::default(),
+            top_line: None,
         })
     }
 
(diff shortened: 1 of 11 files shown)


src/ui/split_screen.rs
```diff
@@ -682,32 +689,65 @@ impl SplitScreen {
...
-            tags.sort();
-            let tags = tags.join("");
-            let mut output = format!("{host}{tags}");
-            if !output.is_empty() {
-                output.push(' ');
-            }
-            write!(self.screen, "{:═<1$}", output, self.width as usize)?; // Print separator
+            let mut default_output = String::default();
+            let output = self.top_line.as_ref().unwrap_or_else(|| { default_output = self.default_top_bar(); &default_output });
+            Self::draw_bar(self.width as usize, 1, &mut self.screen, output)?;
```
krendil: This method turned out a bit ugly and I'm not entirely happy with it, but I'm also a bit of a Rust newbie and couldn't find a nicer way to do things while still: a) Not creating an unnecessary copy of `self.top_line`, and b) only evaluating `self.default_top_bar()` if we need to.

To avoid cloning `self.top_line`, we need to pass it as a reference. This is fine, but that means we need to get the output of `self.default_top_bar()` as a reference as well, which you can't do directly since it is a temporary value. Hence `let mut default_output`.

Since `output` is an immutable reference to `self` that we are passing to `draw_bar`, we can't also pass a mutable reference to self, so we have to pass `screen` and `self.width` as separate parameters instead of grabbing them from `self` inside the method, and call draw_bar as a static method.

src/ui/split_screen.rs
```diff
@@ -682,32 +689,65 @@ impl SplitScreen {
...
+        tags.sort();
+        let tags = tags.join("");
+        let mut output = format!("{host}{tags}");
+        if !output.is_empty() {
+            output.push(' ');
+        }
+        return output;
+    }
+
+    // Almost identical to StatusArea.draw_bar, but with a double line bar character
```
krendil: I tried extracting this and StatusArea::draw_bar into a common function, but unfortunately, the format fill character needs to be known at compile time, and the status area and the top bar use different bar characters.

krendil:
This change scratches a personal itch for me, in that I'd effectively like a second status area to display infrequently changing info, like character name and level, and save the bottom status area for more important and frequently changing info, like room exits and character vitals.

I couldn't figure out how to test this functionality, and I can't see any existing tests for SplitScreen (the ones in split_screen.rs seem to be testing History instead), but I'm happy to add tests if someone points me in the right direction.

I understand if you prefer to keep the top bar for its current purpose and decline to merge this PR.

LiquidityC:
I think this looks pretty good based on the code. Can you post some screenshots of the change in this PR?

krendil:
Passing `nil` will reset to default top bar:
Screenshot from 2026-05-27 08-42-25

Passing empty string will result in unbroken top bar:
Screenshot from 2026-05-27 08-42-44

Text is incorporated into bar:
Screenshot from 2026-05-27 08-42-52

Colours are reset after custom line content (only just fixed the bg not being reset!):
Screenshot from 2026-05-27 08-58-23

Example use case (setup is still WIP):
Screenshot from 2026-05-27 08-45-46

src/ui/split_screen.rs
```diff
@@ -682,32 +689,69 @@ impl SplitScreen {
...
+            &String::default() // Empty String
+        };
+        let mut tags = self
+            .tags
+            .iter()
+            .map(|s| format!("[{s}]"))
+            .collect::<Vec<String>>();
+        tags.sort();
+        let tags = tags.join("");
+        let mut output = format!("{host}{tags}");
```
LiquidityC: The space between `{host}` and `{tags}` should be retained.

LiquidityC: The space between `{host}` and `{tags}` should be retained.

src/ui/split_screen.rs
LiquidityC: Try and make a common free function with as much as possible. The leading `=` or `-` can be provided as an argument and the tail can be printed outside of the function maybe?

I'd be ok with replacing the fancy formatting/padding logic with something more readable, like building a str with a repeated char.

Instead of making a `Line` object below to get the clean line length you might be able to leverage the printable_chars iterator for len.

src/ui/split_screen.rs
krendil: This should be fixed in the latest commits

src/ui/split_screen.rs
krendil: I've made the changes you suggest, pulling draw_bar out into a single common function, using PrintableCharsIterator to get the display width, and building a temporary string for the end of bar padding characters.

I've also made draw_bar truncate the text so that it fits within the bar (including allowing room for the trailing space and bar character at the end). Note that this will affect bar lines in the status area as well, but not non-bar lines. Previously, the top bar would be truncated at the end of the line (without leaving space for the end of the bar), and status area bar lines would overflow onto the next line.

I'm not able to update the screenshots at the moment, I'll do so when I have some time tomorrow.

krendil:
It looks like there might be some conflicts with #1444. Conflicts in `redraw_top_line()` will need to be resolved and the `top_line` help file from this PR should probably be merged into the `top_area` help file from that one.

I'm not sure what the best approach is, is there anything that can be done ahead of time on this branch to make the merge easier?

src/ui/split_screen.rs
LiquidityC: No worries. I don't need more screenshots.

kitsumiko:
Hi @krendil — thanks for flagging the overlap on #1444. I've taken a swing at integrating both directions on the `feat/tabs` branch:

- Your 6 commits are cherry-picked verbatim onto `feat/tabs` with `git cherry-pick -x`, so your authorship is preserved on those commits in the PR history (they show up as authored by you with provenance trailers back to this PR's SHAs).
- On top of those, a refactor commit extracts the top-area rendering into a new `src/ui/top_area` module that organizes the bar(s) as an ordered list of addressable rows. Your `top_line` API survives as the canonical Lua sugar — it just routes through the row system now instead of being a special-case field on `SplitScreen`.
- A second feature commit adds `set_top_row` / `add_top_row` / `remove_top_row` so scripts can configure additional rows (e.g. character vitals, room exits) using the same primitive your `top_line` uses.
- Your standalone `top_line.md` is folded into a rewritten `top_area.md` per your suggestion above, with `settings.md` re-pointed accordingly.

Two paths forward, both fine with me:

1. **#1444 lands first** — `top_line` is already in the modular system, so you could close this as merged-via-#1444 with your authorship visible in the git history.
2. **#1436 lands first** — `feat/tabs` rebases onto post-merge `dev`, the cherry-picks drop out as duplicates, and only my two refactor/feature commits replay.

Tagging @LiquidityC for ordering preference. Happy to adjust the approach if either of you would prefer something different.

LiquidityC:
This PR is very close to merge. Just need to check one thing. Once it's merged the other PR can rebase on dev

LiquidityC (approved):

PR #1439, opened 2026-05-26
Add log_timestamps setting and fix GBK codec encoding

llps112:
This PR contains two improvements:

## 1. Add log_timestamps setting

Adds a new `log_timestamps` boolean setting that, when enabled, prefixes each session log line with a timestamp in `[HH:MM:SS]` format.

- New setting: `/set log_timestamps <on/off>`
- Default: false (disabled)
- Takes effect immediately without restart
- Only affects new log lines, existing log files unchanged
- Updated help documentation and completions

**Files modified:**
- `src/model/settings.rs` - Add LOG_TIMESTAMPS constant and setting
- `src/io/logger.rs` - Add timestamps field and set_timestamps() method
- `src/session.rs` - Pass log_timestamps setting to Logger
- `src/lib.rs` - Handle SettingChanged event for log_timestamps
- `resources/help/logging.md` - Add timestamps documentation
- `resources/help/settings.md` - Add log_timestamps to settings list
- `resources/completions.txt` - Add log_timestamps to completions
- `resources/lua/types/blightmud.d.lua` - Update type definitions
- `docs/log_timestamps_feature.md` - Feature documentation

## 2. Fix GBK codec encoding when sending text to server

The `--codec` option only handled the receive direction (server→client: GBK → UTF-8), but missed the send direction (client→server: UTF-8 → GBK). This caused Chinese characters typed by players to appear garbled on GBK-encoded MUD servers.

**Fix:** In `src/event.rs`, when handling `Event::ServerInput`, if a codec is configured, encode the output buffer from UTF-8 to the target encoding before sending.

**Files modified:**
- `src/event.rs` - Add encoding conversion in send path


LiquidityC:
Please submit this as two PRs. One per change.

PR #1441, opened 2026-05-27
Add log_timestamps setting to prefix session log lines with timestamps

llps112:
Adds a new `log_timestamps` boolean setting that, when enabled, prefixes each session log line with a timestamp in `[HH:MM:SS]` format.

- New setting: `/set log_timestamps <on/off>`
- Default: false (disabled)
- Takes effect immediately without restart
- Only affects new log lines

**Files modified:**
- `src/model/settings.rs`
- `src/io/logger.rs`
- `src/session.rs`
- `src/lib.rs`
- `resources/help/logging.md`
- `resources/help/settings.md`
- `resources/completions.txt`
- `resources/lua/types/blightmud.d.lua`
- `docs/log_timestamps_feature.md`

PR #1443, opened 2026-05-27
Ansi color reset

llps112:
## Problem

When MUD servers send ANSI-colored text without a trailing reset sequence (e.g. `\e[35m82084450` without `\e[0m`), the terminal retains the color state and subsequent lines inherit it, causing visual corruption.

For example, if the server sends purple text for experience points without resetting the color, the next line (including the prompt) will also appear purple.

## Root Cause

The following functions in `src/ui/split_screen.rs` render MUD output directly without ensuring ANSI color state is reset:

- `render_history_line()` - Used for history display during scrolling
- `print_line()` - Used for live output line printing  
- `redraw_prompt()` - Used for MUD prompt redrawing

These functions call `line.print_line()` or `line.tagged_line()` which return the raw ANSI content from the MUD server. When this content doesn't end with `\e[0m`, the terminal color state persists to the next rendered element.

## Solution

Append `\e[0m` (ANSI reset) after rendering each line in all three functions. This ensures the terminal color state is always reset after each rendered line, preventing color leakage to following content.

The fix is safe because:
- ANSI reset is idempotent - calling it multiple times has no negative effect
- Lines that already end with `\e[0m` are unaffected
- This is a standard practice in terminal applications

## Files Modified

- `src/ui/split_screen.rs` - Added `\e[0m` reset in 3 rendering functions
ANSI_COLOR_PR


LiquidityC:
This PR contains the same changes as in #1441 in addition to the described changes.

However I won't be approving this PR. It is the servers job to correctly control colors and properly reset them. Many muds depend on colors being set over multiple lines. So if you are seeing this issue with a mud the correct fix is to report it to the admins/maintainers of that mud.
You can also solve this for particular cases through scripting. Just print a <reset> before printing your prompt.

llps112:
> This PR contains the same changes as in #1441 in addition to the described changes.
> 
> However I won't be approving this PR. It is the servers job to correctly control colors and properly reset them. Many muds depend on colors being set over multiple lines. So if you are seeing this issue with a mud the correct fix is to report it to the admins/maintainers of that mud. You can also solve this for particular cases through scripting. Just print a before printing your prompt.

You're absolutely right, I didn't think it through completely. Some MUDs do rely on colors carrying over multiple lines, and forcing a reset would break that behavior. I'll handle this on the Lua scripting side instead — just print \x1b[0m before the prompt when needed. Thanks for the explanation!

PR #1444, opened 2026-05-27
Add scrollable tabbed output regions

### LINKED ISSUE #1432: Feature proposal: scrollable tabbed output regions (blight.create_tab, F-key switching, filter routing)
## Problem

When playing on a busy MUD, a single scrollback buffer mixes chat, combat, and ambient room output. Catching up on chat after a long fight means scrolling back through hundreds of unrelated lines, and pausing the screen to read chat interrupts combat. A common request in MUD clients (TinTin++ has `#tab`, Mudlet has miniconsoles) is named, independently scrollable output regions you can swap between with a keypress.

I prototyped this for Blightmud on a fork branch and would like to gauge maintainer interest before opening a PR. The feature is additive — every existing script keeps working unchanged — and the implementation is Lua-script-driven, so users who don't use it never see it.

## Proposed Lua API

```lua
-- Create a named tab with optional display label + key-hint
blight.create_tab("chat", { label = "chat", shortcut = "F2" })
blight.create_tab("combat", { label = "combat", shortcut = "F3", gag_main = true })

-- Route lines into a tab by regex (compiled once, OR-d together)
blight.add_tab_filter("chat", "^[A-Z]\\w+ tells you")
blight.add_tab_filter("chat", "^<\\w+> [A-Z]")

-- Switch the active tab (bind to a key)
blight.bind("f2", function() blight.switch_tab("chat") end)

-- Live label / shortcut updates + direct output
blight.set_tab_label("chat", "chat (afk)")
blight.set_tab_shortcut("main", "F1")
blight.output_to("combat", "BOSS DOWN: " .. mob_name)

-- Read-only introspection (for script-driven status lines)
for _, t in ipairs(blight.tabs()) do
  print(t.name, t.label, t.unread, t.active)
end
local active = blight.active_tab()
```

`main` is an implicit reserved tab that always exists. Routing semantics:

- **Mirror by default.** A matched line appears in both `main` and the tab. `gag_main = true` opts a tab into "move" semantics (the line routes to the tab only).
- **Per-tab scroll position** is preserved across switches.
- **Unread counter** per non-active tab; clears on switch.

## Indicator visuals

When two or more tabs are configured, an indicator renders at the top of the screen. Two layouts, controlled by a setting:

Dedicated row (default — `tab_indicator_inline=false`):

```
═══ Blightmud ══ [F1 - main] │ (F2 - chat·3) │ (F3 - combat) ═════
═ host:port [tags] ════════════════════════════════════════════════
```

Inline (shares the host topbar row, `tab_indicator_inline=true`):

```
═ host:port [tags] [F1 - main] │ (F2 - chat·3) │ (F3 - combat) ═══
```

- Active tab: bracketed in bold light-green
- Inactive with unread: yellow parens + `·N` count
- Inactive idle: dim parens
- Brand prefix `═══ Blightmud ══` is configurable (`tab_indicator_brand`)

When only the implicit `main` tab exists, the indicator hides itself — the layout is bit-identical to stock Blightmud.

## Compatibility

- **Pure opt-in.** Scripts using only `blight.output(...)` are unchanged. The tab indicator and routing kick in only after the first `blight.create_tab()` call.
- **No new settings are persisted unless touched.** `settings.ron` only gains entries when `/set` is used or a script changes them via Lua.
- **All new Lua functions are additive** under `blight.*`. No existing function changes signature.
- **Triggers** are independent of tab filters — both passes run on the same line without coordination. Trigger callbacks can use the existing `blight.output_to(tab, line)` for conditional routing.

## Why this design

- **Snapshot-based screen API.** The screen owns no `Arc<Mutex<TabSet>>`. After each TabSet mutation, the event handler calls `screen.set_tab_indicator(tab_set.list())` with a fresh `Vec<TabInfo>`. Keeps the UI layer free of lock dependencies.
- **History swap on switch.** The active tab's `History` lives in the screen; inactive tabs hold theirs inside `Tab.history`. Switching is a three-step take/swap/complete protocol — no scrollback copy.
- **Regex filters are compiled once, evaluated single-pass per line.** `tab_set.route()` returns a `RouteResult { render_to_screen, indicator_dirty }` so the hot path knows whether to skip rendering (e.g. when a line was gagged into a tab) or refresh the indicator.
- **Indicator is rendered from cached metadata.** No per-line allocation or lock when no inactive tabs are matched.

## Implementation status

A working implementation lives at: **https://github.com/kitsumiko/Blightmud/tree/feat/tabs**

11 commits on top of `origin/dev`, ~1655 line additions across 20 files:

```
 resources/help/{settings,tabs,top_area}.md  (3 files: docs)
 src/event.rs                                 (TabCommand + handler arms)
 src/lib.rs                                   (hot-reload arm)
 src/lua/{blight,lua_script}.rs               (Lua API)
 src/model/settings.rs                        (3 new settings)
 src/session.rs                               (TabSet ownership)
 src/tabs/{mod,tab,tab_set}.rs                (new module)
 src/ui/{headless,reader,split,ui_wrapper}.rs (UserInterface impls)
 src/ui/{help_handler,mod,user_interface}.rs  (trait + help registration)
```

The screen-side renderer is `SplitScreen`-only; `ReaderScreen` and `HeadlessScreen` get no-op implementations of the new trait methods.

`cargo test` (existing + new TabSet unit tests) passes; built with `--release` in a clean container under qemu-emulated targets without warnings beyond pre-existing ones.

## Open questions before I send a PR

1. **Module placement** — I put the data layer in a new `src/tabs/` module. Would you prefer `src/model/tabs.rs` (single file, alongside `Line`, `Settings`, etc.) or a `model::tabs` submodule?
2. **Settings count** — three new settings (`tab_indicator_visible`, `tab_indicator_brand`, `tab_indicator_inline`). Acceptable, or would you rather collapse to one string-valued setting (e.g. `tab_indicator = "off" | "row" | "row-no-brand" | "inline"`)? The current `Settings` HashMap is bool-only so a string value would need a parallel mechanism.
3. **`MAIN_TAB` constant exposure** — currently `pub const` inside `tabs::tab`. Should Lua scripts be able to read it (e.g. `blight.main_tab_name()`), or is the string `"main"` enough?
4. **Indicator branding** — the default indicator includes a bold `═══ Blightmud ══` prefix. Toggle-off via `tab_indicator_brand=false` is supported. Acceptable as a default-on visual, or should it default off?
5. **Help topic scope** — `/help tabs` documents the Lua API, `/help top_area` documents the screen region. The latter is new and parallels `/help status_area`. Is the two-topic split worth it?
6. **Scope of this PR** — should I split this into smaller PRs (e.g. core tab routing + indicator + Lua API as separate reviews), or submit as a single feature PR?

Happy to adjust any of these before opening the PR. If the overall direction is something you'd want in Blightmud, I'll send the PR against `dev` once we've aligned on the points above.

If not — totally understood; it lives on a fork branch and isn't blocking anything.

kitsumiko:
**Builds on #1436.** Six of the commits in this branch are cherry-picked from @krendil's `custom-topbar` branch (#1436) with `-x` provenance trailers — they appear authored by him in the PR commit list. The subsequent `refactor(top-area)` and `feat(top-area)` commits extract a modular `TopArea`/`TopRow` system that subsumes both PRs' top-area work: `blight.top_line` from #1436 becomes one-line sugar for the new `blight.set_top_row("host_status", { body = ... })` primitive, and inline-tab mode composes orthogonally with custom top-line content.

---

Closes #1432.

Implements the scrollable tabbed output regions feature proposed in #1432, plus the `add_tab_exclude_filter` follow-up flagged in the issue thread (the lookaround gap that hits as soon as you write a broad regex). Pure opt-in — scripts using only `blight.output(...)` are unchanged, and the indicator row hides when only the implicit `main` tab exists, so the default layout is bit-identical to stock Blightmud.

## Lua API

```lua
blight.create_tab("chat", { label = "chat", shortcut = "F2" })
blight.add_tab_filter("chat", "^[A-Z]\\w+ tells you")
blight.add_tab_exclude_filter("chat", "^(He|She) says\\b")  -- carve exceptions
blight.bind("f2", function() blight.switch_tab("chat") end)
blight.output_to("combat", "BOSS DOWN")
for _, t in ipairs(blight.tabs()) do print(t.name, t.unread, t.active) end
```

Full reference in `/help tabs`.

## Implementation

- **New module** `src/tabs/{mod,tab,tab_set}.rs` — data layer (`TabSet`, `Tab`, `TabOpts`, `TabInfo`). 11 unit tests cover filter matching, exclude veto, gag-main suppression, unread bumping, and the switch protocol.
- **Routing**: `EventHandler::route_and_print()` calls `tab_set.route(&line)`. Inactive tabs accumulate matched lines in their own `History`; the active tab's `History` lives in the screen and is swapped on `switch_tab()`. No `Arc<Mutex<TabSet>>` reaches the screen.
- **Indicator**: `SplitScreen` renders a configurable header row gated by `tab_indicator_visible` / `tab_indicator_brand` / `tab_indicator_inline`. `ReaderScreen` and `HeadlessScreen` get no-op trait stubs.
- **Lua surface**: 12 additive methods on `blight.*`; no existing function changes signature.

## Files modified

- new module: `src/tabs/{mod,tab,tab_set}.rs`
- ui: `src/ui/{split_screen,ui_wrapper,user_interface,headless_screen,reader_screen}.rs`
- session / event / lua: `src/{event,lib,session}.rs`, `src/lua/{blight,lua_script}.rs`
- settings: `src/model/settings.rs` (3 new bool settings)
- docs: `resources/help/{tabs,top_area,settings,help}.md`

No changes to `Cargo.toml`, `Cargo.lock`, or `build.rs`.

## Test plan

- `cargo test --release --lib`: 395 pass (384 existing + 11 new `tab_set` tests)
- `cargo build --release`: clean
- Live smoke test on SWmud: tab switching, filter routing, exclude veto, gag_main, inline-vs-row indicator toggle all behave per the docs

## Open questions from #1432

These remain open if you'd prefer a different shape — happy to adjust:

1. **Module placement** — `src/tabs/` vs `src/model/tabs.rs`
2. **Settings count** — three bools vs one string-valued `tab_indicator` enum
3. **`MAIN_TAB` exposure** — currently `pub const` only, no Lua getter
4. **Brand default** — `═══ Blightmud ══` prefix is on by default; toggle via `tab_indicator_brand=false`
5. **Help split** — `/help tabs` + `/help top_area` (two topics) vs one
6. **PR scope** — sending as one PR; happy to split into core / indicator / Lua-API if preferred


kitsumiko:
## Update: integrated @krendil's #1436 into this branch

Following @krendil's comment on #1436 about the conflict, I've pulled both PRs' concepts into `feat/tabs` as a single modular system:

**Commits added (in order):**
```
krendil  Allow customizing the content of the top bar        (-x: 6e9833d)
krendil  Run cargo fmt                                        (-x: 537598c)
krendil  Reset background colour when drawing custom top line (-x: 531d50f)
krendil  Restore space between host and tags in default top line (-x: 6a6895d)
krendil  Refactor draw_bar into a single standalone function  (-x: 8a6d458)
krendil  Truncate text to fit within line width               (-x: 10fde62)
me       refactor(top-area): extract TopArea/TopRow modular system
me       feat(top-area): set_top_row / add_top_row / remove_top_row Lua API
me       style: cargo fmt + collapse identical if branches in render_row
```

All six of @krendil's commits are cherry-picked verbatim with `-x` provenance trailers — they're authored by him in the commit list, not squashed or co-author-trailed.

**What the modular system gives you:**

The top area is now an ordered list of addressable `TopRow`s. Each row is `{ name, bar_char, prefix, body, visible }`. Two built-ins ship by default — `tab_indicator` (this PR) and `host_status` (the `host:port [tags]` bar, body-overridable via #1436's `blight.top_line`).

```lua
blight.top_line("HP 80/100")                          -- #1436's API, now sugar for…
blight.set_top_row("host_status", { body = "HP 80/100" })  -- …the generic primitive

blight.add_top_row({                                  -- NEW: script-defined rows
  name = "vitals", bar_char = "─",
  prefix = { text = " Vitals ", style = "brand" },
  body = "",
})
blight.remove_top_row("vitals")
```

Inline-tab mode (`/set tab_indicator_inline true`) composes orthogonally with custom top-line content — when both are set, tabs splice in after whatever's in `host_status.body`, so you keep your character vitals AND your tab strip on the same row.

**Other changes:**
- `resources/help/top_area.md` rewritten as the canonical reference, folding in #1436's `top_line` docs per @krendil's [own suggestion](https://github.com/Blightmud/Blightmud/pull/1436#issuecomment-4560005194).
- `resources/help/top_line.md` deleted (folded).
- `resources/help/settings.md` `hide_topbar` entry re-points at `/help top_area`.
- `resources/lua/types/blightmud.d.lua`: stubs for `set_top_row` / `reset_top_row` / `add_top_row` / `remove_top_row` / `top_rows`.

**Test counts:** 384 (pre-#1444) → 395 (#1444 alone) → 398 (after cherry-picks add krendil's 3 `draw_bar` tests) → **409** (after the refactor adds 11 `top_area::tests`).

Earlier validate failure was just `cargo fmt` drift on the new module — fixed in fd1555d.

kitsumiko:
Rebased onto current `dev` now that #1436 has merged.

Since @krendil's top-bar customization (#1436) is in `dev`, I dropped the six cherry-picks of his commits this branch had been carrying (now redundant) and folded the interim tab-indicator iterations into the single `TopArea`/`TopRow` refactor that unifies both features. The net top-area behavior is unchanged: `host_status` (the #1436 bar) and `tab_indicator` are the two built-in rows, `blight.top_line(...)` now routes through `set_top_row("host_status", ...)`, and `top_line.md` is folded into `top_area.md`.

Branch is now three commits on top of `dev`:
- `feat(tabs): scrollable named output regions with Lua API + key-bindable switching`
- `feat(tabs): add_tab_exclude_filter — blocklist-over-include routing`
- `feat(tabs+top-area): tab indicator + unified TopArea/TopRow system` (co-authored with @krendil)

Conflicts resolved; `cargo test` is green (409 unit tests + the integration suites) against dev's current dependencies.

kitsumiko:
Added test coverage for the tab / top-area code — coveralls is green again (*Coverage increased (+1.6%) to 75.9%*).

The feature added a lot of new surface that wasn't exercised by tests, which dropped overall coverage. Now covered by:
- `tests/tab_tests.lua` (wired into `tests/lua_tests.rs`) — a headless integration test driving the full Lua API: `create_tab` / `switch_tab` / `add_tab_filter` / `add_tab_exclude_filter` / `set_tab_label` / `set_tab_shortcut` / `output_to` / `tabs` / `active_tab`, plus the `set_top_row` / `add_top_row` / `remove_top_row` / `reset_top_row` / `top_rows` top-area API.
- Unit tests for `tabs/tab.rs`, `tabs/tab_set.rs`, and `ui/top_area.rs` (routing, label/shortcut/output_to, and `TopRow` resolve/apply/add/remove/render).

No production code changed.

LiquidityC:
It's late but I've started looking into this PR

I need to test some things and it will be a while because it's large. I'd like to ask what the expected additional memory consumption is for a tab? Is there a benefit to controlling this per tab? Perhaps tabs should or can have shorter histories?

kitsumiko:
Thanks for taking a look — this turned out to be worth chasing, good call.

**What a tab cost before**

Each secondary tab owns a `History`, and `History::new()` eagerly calls `Vec::with_capacity(32 * 1024)` twice (`inner` + `visible`). With `size_of::<Line>() == 152` (Linux) that's **9.5 MiB reserved the moment a tab is created**, before any output, paid per tab whether it's used or not:

| tabs | 1 | 5 | 20 | 50 |
|---|---|---|---|---|
| reserved | 9.5 MiB | 47.5 MiB | 190 MiB | 475 MiB |

dhat confirms it to the byte: `History::new()` → +9,961,472 B (= 2 × 32768 × 152).

**Change I pushed**

1. Tab histories are now **lazy** — the Vecs grow on demand instead of being reserved up front, same drain ceiling so scrollback depth is unchanged.
2. Optional **`history_lines`** on `create_tab` to cap a tab, e.g. `create_tab("combat", { history_lines = 2000 })`.

Measured after, headless on Linux (peak RSS):

| | baseline | 50 empty tabs | 50 tabs × 1000 lines |
|---|---|---|---|
| RSS | 9.7 MiB | 9.7 MiB (+~20 kB) | 44.7 MiB (+~35 MiB of content) |

Directly answering:

- **Expected memory per tab:** ~0 when idle (dhat: an empty tab is **+8 B**, just the name), growing only with the lines it actually holds. Empty tab: 9.5 MiB → 8 B.
- **Per-tab control:** yes, via `history_lines`. A 2000-line tab filled past its ceiling peaks at **~0.94 MiB** vs 9.5 MiB (16.5×).
- **Shorter histories:** supported by `history_lines`; and with lazy growth even a full-depth tab costs nothing until used.

No leaks: a create/drop loop returns to baseline under dhat (0 retained), and valgrind reports 0 bytes definitely/indirectly lost.

One related note: the same eager `History::new()` backs the **main** scrollback in a normal session — I left it alone to keep this PR focused, but making it lazy too would trim ~9.5 MiB off the baseline. Happy to fold that in here or as a follow-up, your call.

Full measurement harness (size analysis + headless RSS sampler + dhat probes, one-command repro in the commit message) is on a branch off this one if it's useful: [`feat/tabs-mem-bench`](https://github.com/kitsumiko/Blightmud/tree/feat/tabs-mem-bench).

kitsumiko:
Quick CI note: the one red check, **Tests (ubuntu-latest, beta)**, is a flaky segfault — the lib-test binary exits with SIGSEGV (signal 11) in the `--all-features` run, right after `lua::spellcheck::tests::test_check` (C FFI). The same suite passed on the other five matrix cells (including macOS-beta) and locally, so it's unrelated to the tab changes.

Would you mind re-running that job when convenient? I don't have permission to re-trigger CI from the fork. Thanks!

resources/help/tabs.md
```diff
@@ -0,0 +1,315 @@
...
+
+```lua
+-- Set up two tabs alongside main, each with a shortcut hint shown
+-- in the indicator. (Main's shortcut is set after the fact because
+-- "main" is reserved and not created via create_tab.)
+blight.create_tab("chat",   { label = "chat",   shortcut = "F2" })
+blight.create_tab("combat", { label = "combat", shortcut = "F3" })
+blight.set_tab_shortcut("main", "F1")
+
+blight.add_tab_filter("chat",   "tells you|^%S+ shouts|^%S+ chats|^%S+ gossips")
```
LiquidityC: This isn't regex syntax is it?

src/lua/blight.rs
```diff
@@ -54,6 +73,87 @@ impl Blight {
...
+        opts.bar_char = Some(ch);
+    }
+    if table.contains_key("prefix")? {
+        let value: mlua::Value = table.get("prefix")?;
+        opts.prefix = Some(top_prefix_from_lua(&value)?);
+    }
+    if table.contains_key("body")? {
+        let value: mlua::Value = table.get("body")?;
+        opts.body = Some(match value {
+            mlua::Value::Nil => TopRowBody::Empty,
```
LiquidityC: I'm not sure this can be reached. A key with the value 'nil' in lua is the same as no key. Should this be empty string?

src/lua/blight.rs
```diff
@@ -140,6 +240,46 @@ impl UserData for Blight {
...
+            Ok(())
+        });
+        methods.add_function("top_rows", |ctx, _: ()| -> LuaResult<Table> {
+            // Built-in row names are always present and stable. Lua-added
+            // rows are tracked by the script itself.
+            let arr = ctx.create_table()?;
+            arr.push("tab_indicator")?;
+            arr.push("host_status")?;
+            Ok(arr)
+        });
```
LiquidityC: Should we perhaps call this `builtin_top_rows()` since there's nothing dynamic here?

LiquidityC (commented):
Have 3 small comments that need fixing. Also I'd either like it to be possible to do ìght.tab_remove/delete` or it needs to be documented that there's no way out of a created tab. I'd prefer if it was possible to remove them.

resources/help/tabs.md
kitsumiko: Good catch — that's a docs bug, not valid regex. `%S` is Lua-pattern syntax, but tab filters compile through the Rust `regex` crate (`TabSet::add_filter` → `regex::Regex::new`), so it should be `\S`, written `\\S` in a Lua string literal — the same escaping the `\\b` / `\\w` examples elsewhere in this file already use. Fixed both occurrences (lines 66 and 290) to `^\\S+`.

src/lua/blight.rs
kitsumiko: You're right, it's unreachable — `contains_key("body")` is false when `body` is `nil`, since a nil-valued key doesn't exist in Lua, so the `Value::Nil` arm is dead. Switched to the empty-string sentinel you suggested: `body = ""` ⇒ `Empty` (bar only), non-empty ⇒ `Text`. That also lines up with the existing `top_line("")` "unbroken bar" behavior. I applied the same fix to the sibling `prefix` field, which had the identical dead `nil` arm (`top_prefix_from_lua`): `prefix = ""` now clears it. Updated the `top_row_opts_from_lua` doc-comment and the `/help top_area` field reference to match.

src/lua/blight.rs
kitsumiko: Agreed — renamed to `builtin_top_rows()`. Updated the binding, the `blightmud.d.lua` type stub, `/help top_area`, and the integration test.

kitsumiko:
Thanks for the review @LiquidityC! Pushed a round that addresses all three inline comments plus the tab-removal request. Rundown:

**1. `tabs.md` filter example "isn't regex syntax" (`r3408308105`)** — Correct, that was a docs bug: `%S` is Lua-pattern syntax that crept into a Rust-`regex` example. Tab filters compile via `TabSet::add_filter` → `regex::Regex::new`, so the right form is `\S`, escaped as `\\S` in a Lua string literal (matching the `\\b` / `\\w` examples already in the file). Fixed both filter examples (lines 66 and 290), and grepped the rest of the tabs/top-area docs for stray Lua-pattern classes in regex contexts — these two were the only offenders.

**2. Unreachable `body = nil` arm (`r3408315408`)** — Spot on: `contains_key("body")` is false when the value is `nil` (nil-valued keys don't exist in Lua), so the `Value::Nil` arm was dead. Replaced it with the empty-string sentinel you suggested — `body = ""` ⇒ `Empty` (bar only), non-empty ⇒ `Text` — which also matches the existing `top_line("")` "unbroken bar" behavior. Applied the same fix to the sibling `prefix` field, which had the identical dead `nil` arm (`top_prefix_from_lua`): `prefix = ""` now clears it. Updated the `top_row_opts_from_lua` doc-comment, the `/help top_area` field reference, and the type stubs so the docs no longer advertise the unreachable `nil` behavior.

**3. `top_rows()` → `builtin_top_rows()` (`r3408318171`)** — Renamed; it only ever returns the static built-ins, so the clearer name is right. Updated the binding, the `blightmud.d.lua` stub, `/help top_area`, and the integration test.

**Tab removal (review summary)** — Added `blight.remove_tab(name)`, named to match `create_tab` / `switch_tab`. It refuses `main` (reserved) and unknown tabs with the usual error line; removing the currently-active tab switches back to `main` first (reusing the existing history-swap protocol), then drops the tab and discards its scrollback. Covered by new `TabSet::remove` unit tests and an end-to-end case in `tab_tests.lua`, and documented in `/help tabs`. While wiring this up I noticed the tab Lua API was missing from `blightmud.d.lua` entirely (only the `top_row*` stubs were present), so I added the full set of tab stubs rather than leave `remove_tab` as a lone orphan.

Happy to switch the name to `tab_remove` if you'd prefer, or have active-tab removal fall back to the adjacent tab instead of `main` — let me know.

PR #1448, opened 2026-06-01
Optimize history drain

llps112:
## Summary

Two related optimizations to address RSS growth from malloc fragmentation in long-running sessions:

1. **Incremental drain**: Change `History::drain()` from full `rebuild_visible()` to incremental visible line removal, avoiding repeated 16MB Vec allocation/deallocation cycles
2. **Configurable capacity**: Add `blight.history_capacity()` Lua API for runtime scrollback buffer size control (default: 32768 lines, backward compatible)

## Motivation

- `History::drain()` previously called `rebuild_visible()` which caused large Vec allocation/deallocation cycles
- malloc couldn't reclaim these fragments, leading to continuous RSS growth over time
- The hardcoded 32768 line capacity couldn't be adjusted for long-running sessions

## Changes

- `History::set_capacity()` - dynamic buffer resizing with line trimming
- `History::drain()` - incremental visible line removal instead of full rebuild
- `blight.history_capacity([capacity])` - new Lua API for getting/setting buffer size
- `Event::SetHistoryCapacity` - capacity change event propagation through all UI components
- 3 new unit tests for `set_capacity` functionality
- Updated help documentation

## Testing

- All 22 history-related tests pass
- `cargo fmt` clean
- `cargo clippy` clean (no new warnings from changed files)

## Usage

```lua
-- Get current capacity
local cap = blight.history_capacity()
print("Current capacity: " .. cap)

-- Set new capacity (e.g., 5000 lines for long-running sessions)
blight.history_capacity(5000)
```

### DIFF
--- a/src/ui/history.rs
+++ b/src/ui/history.rs
@@ -37,14 +37,18 @@ impl History {
 
     pub fn set_capacity(&mut self, new_capacity: usize) {
         self.capacity = new_capacity;
-        while self.inner.len() > self.capacity {
-            let drained = self.inner.remove(0);
-            if !drained.is_masked(&self.tag_mask) {
-                if !self.visible.is_empty() {
-                    self.visible.remove(0);
+        if self.inner.len() > self.capacity {
+            let excess = self.inner.len() - self.capacity;
+            let drained: Vec<Line> = self.inner.drain(0..excess).collect();
+            let visible_drain_count = drained
+                .iter()
+                .filter(|l| !l.is_masked(&self.tag_mask))
+                .count();
+            if visible_drain_count > 0 {
+                if visible_drain_count <= self.visible.len() {
+                    self.visible.drain(0..visible_drain_count);
                 } else {
                     self.rebuild_visible();
-                    return;
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


src/ui/history.rs
```diff
@@ -35,10 +35,35 @@ impl History {
...
+                }
+            }
+        }
+    }
+
     pub fn drain(&mut self) {
         if self.inner.len() >= self.capacity {
-            self.inner.drain(0..self.drain_length);
-            self.rebuild_visible();
+            let drained: Vec<Line> = self.inner.drain(0..self.drain_length).collect();
```
LiquidityC: You need to clamp here. `self.inner.drain(0..self.drain_length.min(self.capacity))` or it can overreach now that capacity is user editable.

Please check any other uses of drain_length to ensure this is checked.

src/ui/history.rs
```diff
@@ -35,10 +35,35 @@ impl History {
...
+                if !self.visible.is_empty() {
+                    self.visible.remove(0);
+                } else {
+                    self.rebuild_visible();
+                    return;
+                }
+            }
+        }
+    }
+
```
LiquidityC, line 56: I think this method could leverage the exact same MO as `drain()` below. It would perform better then this loop.

LiquidityC (changes requested):
Looks good, but there's a case where the user can create a crash by accident which I'd like to eliminate before approving this.

llps112:
Got it, I'll think them over.​

llps112:
@LiquidityC  I've addressed your review comments:

- Added `drain_length.min(self.inner.len())` clamp to prevent panic when user sets a very small `history_capacity`
- Rewrote `set_capacity()` to use batch `drain()` instead of while loop, matching the same pattern as `drain()` for better performance
- Added 5 new tests covering edge cases: `capacity=0`, `capacity=1`, drain length clamp with small capacity, visible/inner consistency after set_capacity with mask, and Lua API test for `history_capacity` getter/setter + event dispatch

All 24 history tests pass. Ready for another look.

LiquidityC (approved):

PR #1452, opened 2026-06-08
Use a temporary list to avoid modifying a list being iterated over

jhall11:
Try to prevent undefined behavior when adding/removing triggers and task from within triggers and tasks by creating a temporary list to iterate over.


LiquidityC:
This bug is a valid one. However, copying all the triggers in `check_line` can get seriously expensive. Pretty sure I have like 300+ triggers and `check_line` runs a lot.

I think the better solution here is to track a `self.busy` flag (like a mutex) in the TriggerGroup. Then in `TriggerGroup:add()` check the `self.busy` flag. If true add the trigger to a temporary list, eg `self.pending_triggers`. At the end of `check_line` when the loop is done you can attach whatever is in `self.pending_triggers` to the end of `self.triggers` and clear out `self.pending`.

Did that make sense?

Thanks for finding and tackling this.

Edit: PS. I will request a test for this. So you may get ahead of me by writing one. It will also help you confirm that your logic is sound. Thanks again.

resources/lua/tasks.lua
```diff
@@ -52,39 +54,45 @@ function Task:start()
...
 end
 
 function Task:send(value)
     self.sentData[#self.sentData + 1] = value
 end
 
 function Task:sleep(time)
-    if tasks[self].idle then
+    local data = tasks[self] or pendingTasks[self]
+    if not data or data and data.idle then
```
jhall11, line 79: I am avoiding an "attempt to index a nil value" error here if the task was already removed. not sure if we wanted to silently ignore or throw an error

resources/lua/tasks.lua
```diff
@@ -52,39 +54,45 @@ function Task:start()
...
+        -- not sure why we don't want to allow adding more of a delay to a task. would allow "run task after x seconds once nothing else is happening"
```
jhall11, line 80: I didn't want to change behavior here, but this seems like it shouldn't hurt to allow

resources/lua/tasks.lua
```diff
@@ -52,39 +54,45 @@ function Task:start()
...
     else
-        tasks[self] = { time = tasks[self].time + time }
+        data = { time = data.time + time }
     end
 end
 
 function Task:idle()
-    tasks[self].idle = true
+    local data = tasks[self] or pendingTasks[self]
+    if data then
```
jhall11, line 93: I am avoiding an "attempt to index a nil value" error here if the task was already removed. not sure if we wanted to silently ignore or throw an error

tests/common.lua
```diff
@@ -1,5 +1,5 @@
-function assert_eq(a, b)
-    assert(a == b, string.format("Assertion failed: '%s' != '%s'", tostring(a), tostring(b)))
+function assert_eq(a, b, msg)
+    assert(a == b, string.format("Assertion failed: '%s' != '%s'%s", tostring(a), tostring(b), msg and ". " .. msg or ""))
```
jhall11, line 2: This seemed helpful but I can remove if not wanted

tests/task_tests.lua
```diff
@@ -0,0 +1,219 @@
...
+    assert_eq(spawn_later_task.dead, true)
+    assert_eq(spawn_later_task.success, true)
+    assert_eq(spawn_later_task.error, nil)
+    assert(test_results.spawn_later_waited, "spawn_later task should not have been spawned immediately")
+    assert(test_results.spawn_later_ran, "spawn_later task should have executed")
+
+    -- Assert self-killed tasks calling sleep or idle
+    assert_eq(self_killed_task.dead, true)
+    assert_eq(self_killed_task.error, nil)
+    --fixme assert(test_results.self_killed_ran_after_killed == false, "task should not run after killed")
```
jhall11, line 213: I think we would have to break from the task if we wanted to prevent it from continuing after calling kill() on itself. not sure if this is worth it? we could also throw a warning or something??

jhall11 (commented):
Sorry it took so long, was on vacation then it took me forever to find consistent reproducible tests for modifying a table while iterating over them.

jhall11:
I'm trying this out with my scripts and I think I broke tasks.sleep somehow

jhall11:
After my last changes I've been running my scripts with no issues. Also ran the lua tests with 1000 iterations and no failures

LiquidityC (approved):

Issue #1472, opened 2026-07-19
Multi-line input area


wisq:
Love the project.  My only wish is that I had more space to type longer messages!

I'd love to have the ability to set the input area to be more than just one line tall — like the 3-line tall input area I had when I used TinyFugue — and then set my up/down keybinds to move up and down through it (rather than through the command history).

Fixed height is fine, but it would be even better if it could be set to auto-expand based on how much text has been input.  (But I'm sure that could also just be done via a `prompt.add_prompt_listener` hook.)

kitsumiko:
I've prototyped this and opened it as two PRs: #1479 (a behaviour-neutral refactor) and #1480 (the feature itself).

Since this is labelled `discussion` and not yet accepted, I'd welcome a steer on the design before either gets a close review — several of the decisions below are policy rather than mechanism, and they're much cheaper to change now than after review.

## What I built

`blight.input_height(n)` sets the input area to N rows (1–10, default 1). With the new `input_auto_expand` setting on, N becomes a *minimum* and the area grows and shrinks with the content. One number, no mode enum — a TinyFugue-style fixed 3-row area is `blight.input_height(3)` with auto-expand off.

Defaults are unchanged: one row, auto-expand off, and the existing single-row horizontal-scroll rendering path is untouched.

## Four decisions I'd like opinions on

**1. Enter always submits; a separate key inserts a row.**

`Ctrl-O` is the primary binding, with `Alt-Enter` and the raw `\x1b\r` / `\x1b\n` forms also bound, retargetable via `set_newline_keys{...}`.

`Ctrl-O` rather than `Alt-Enter` as primary because Alt+Enter does not work on a stock Mac — Terminal.app and iTerm2 both ship with Option producing accented characters rather than a Meta prefix, and some braille/keyboard-emulation stacks never pass it through. `Ctrl-J` isn't usable at all, being byte-identical to Enter. (Worth noting this affects the *existing* `alt-b`/`alt-f`/`alt-d` defaults on stock macOS too — I've documented it but not tried to fix it here.)

Binding `alt-enter` also needed a fix: `Key::Alt('\r')` minted the binding name `"alt-\n"`, a literal newline, so `blight.bind("alt-enter", ...)` silently never fired.

**2. Submit fans out into one `ServerInput` per row.**

[...]

PR #1479, opened 2026-07-22
Extract prompt layout into pure, testable modules

kitsumiko:
Behaviour-neutral refactor, extracted from work on a multi-line input area (#1472). Opening it separately because it stands on its own: it deletes duplicated arithmetic and fixes three reachable scroll-region defects.

> **First of two PRs — merge this one first.**
> #1480 implements the actual feature and is built directly on this branch. It cannot merge until this one does.
> If you'd rather not take the feature at all, this PR still stands on its own.

## Why

`SplitScreen::new` requires a real terminal, so none of the row/cursor arithmetic inside it can run in CI — the test module says so itself:

```rust
// Tests for print_line tag mask behaviour. SplitScreen::new requires a real
// terminal so we exercise the behaviour through History directly...
```

Layering wrap and cursor arithmetic onto that means shipping it untested, in the code path that produced #742. So the arithmetic moves into two pure modules and the renderer becomes a blitter.

## What's fixed

The layout math was duplicated and divergent between `new()` and `setup()` — `new()` hardcoded `output_start_line = 2` while `setup()` derived it. Both now call one `ScreenLayout::compute`. That surfaced three defects:

1. **`height - status_height - 2` underflowed** on short terminals. The near-miss matters more than the panic: saturating it would produce a DECSTBM region with `bottom <= top`, which terminals *discard entirely*, leaving the previous region in force — so every MUD line would scroll the whole screen and drag the status and input areas with it. That reads as "the UI is a bit glitchy" and survives casual testing far better than a crash. `compute` returns `Option` and refuses to render rather than emitting an invalid region.

2. **`init_scroll` emitted `ScrollRegion(scroll_range + 3, ...)`** while drawing its divider at `scroll_range + output_start_line`. The `3` is only correct while `output_start_line == 2`, which stops being true as soon as the tab indicator row appears — the divider then lands *inside* the scroll region and scrolls away on the next line of output.

3. **`scroll_range()` guarded on the terminal height**, which says nothing about how many rows the output region was actually left with after the top, status and input areas took their share, then subtracted `SCROLL_LIVE_BUFFER_SIZE` from it. With fewer than ten output rows this wrapped to ~65530 and was handed to DECSTBM.

Plus two coordinate-space defects, both the bug class that produced #742:

4. **`PromptMask::mask_buffer` advanced its offset by the mask's byte length** while splicing into a `Vec<char>`. ASCII masks hide it; any multi-byte mask from Lua displaced every subsequent index.

5. **`print_prompt_input` received `pos`** — documented one line above as "a character index" — and passed it to `byte_index_at_display_width`, which reads its argument as a *column* count. Those agree only while every character is one column wide, so a CJK glyph left of the cursor dragged it back one column per glyph. Separately, `SetPromptMask` paired the *masked* buffer with an *unmasked* cursor position. Masks are arbitrary Lua strings and printable masks are the documented use (`"this is *important*, ok"`), so `PromptMask::masked_index` now maps between the two spaces.

## What's new

- `src/ui/layout.rs` — vertical screen arithmetic as a pure function. 8 tests.
- `src/ui/input_layout.rs` — `rows`, `cursor_row_col`, `desired_height`, `visible_rows`. 21 tests. `cursor_row_col` is a *lookup into* `rows()`'s partition rather than a parallel reconstruction, because `byte_index_at_display_width` pushes a straddling wide character forward and leaves a hole — an independent calculation disagrees by a full row at every such hole.
- `wrap_line_hard` alongside an untouched `wrap_line`. `wrap_line` is lossy *by design* for MUD output — it drops the word-break space and discards a whitespace-only tail — which is right for prose and wrong for an input area, where the rows must partition the buffer exactly. `wrap_line_hard_preserves_all_characters` asserts both properties side by side so the justification can't quietly stop being true.
- 6 render tests against a capturing writer, which is what finally gets `print_prompt_input` under test without a TTY.

## Not behaviour-neutral

Three deliberate exceptions to the above:

- The `scroll_range` fix enables split-scroll on ~15–20-row terminals where the old height-based guard refused it. That's correct — there *is* room — but it's a change.
- The mask and character-index fixes change rendering for CJK input and for printable masks.
- The `cursor::Save`/`Restore` pair in `print_prompt_input` is gone. DECSC has one save slot per screen buffer, and saving there destroyed the position `setup()` deliberately stores at line 238; the restore was unobservable anyway because `goto_prompt()` overwrites the cursor on the very next byte.

## Testing

495 unit tests and all integration suites pass (`cargo test --locked`). `cargo fmt --check` is clean, and `cargo clippy --locked` produces a warning set identical to `dev` — no new warnings.

The new regression tests were checked by mutation: reverting the erase loop to a content-bounded clear fails `every_input_row_is_erased_even_when_content_shrinks`, and removing the scroll-window offset fails `content_taller_than_the_area_scrolls_within_it`.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -463,9 +463,6 @@ impl EventHandler {
                     lua_ctx.set_prompt_mask_content(updated_mask_table);
                     let mut prompt_input = self.session.prompt_input.lock().unwrap();
                     *prompt_input = command_buffer.get_masked_buffer();
-                    // Masked string, masked cursor. `get_pos` indexes the raw
-                    // buffer and would point into the wrong place as soon as a
-                    // mask inserts anything before the cursor.
                     screen.print_prompt_input(&prompt_input, command_buffer.get_masked_pos());
                 }
                 Ok(())
@@ -478,9 +475,6 @@ impl EventHandler {
                     }
                     let mut prompt_input = self.session.prompt_input.lock().unwrap();
                     *prompt_input = command_buffer.get_masked_buffer();
-                    // The mask was just cleared, so this is the identity — but
-                    // it is spelled the same way as the branch above so the
-                    // pairing stays obviously correct.
                     screen.print_prompt_input(&prompt_input, command_buffer.get_masked_pos());
                 }
                 Ok(())
--- a/src/event.rs
+++ b/src/event.rs
@@ -463,7 +463,10 @@ impl EventHandler {
                     lua_ctx.set_prompt_mask_content(updated_mask_table);
                     let mut prompt_input = self.session.prompt_input.lock().unwrap();
                     *prompt_input = command_buffer.get_masked_buffer();
-                    screen.print_prompt_input(&prompt_input, command_buffer.get_pos());
+                    // Masked string, masked cursor. `get_pos` indexes the raw
+                    // buffer and would point into the wrong place as soon as a
+                    // mask inserts anything before the cursor.
+                    screen.print_prompt_input(&prompt_input, command_buffer.get_masked_pos());
                 }
                 Ok(())
             }
@@ -475,7 +478,10 @@ impl EventHandler {
                     }
                     let mut prompt_input = self.session.prompt_input.lock().unwrap();
                     *prompt_input = command_buffer.get_masked_buffer();
-                    screen.print_prompt_input(&prompt_input, command_buffer.get_pos());
+                    // The mask was just cleared, so this is the identity — but
+                    // it is spelled the same way as the branch above so the
+                    // pairing stays obviously correct.
+                    screen.print_prompt_input(&prompt_input, command_buffer.get_masked_pos());
                 }
                 Ok(())
             }
--- a/src/model/prompt_mask.rs
+++ b/src/model/prompt_mask.rs
@@ -27,10 +27,6 @@ impl PromptMask {
                 return masked_buf.iter().collect();
             }
             masked_buf.splice(adjusted_idx..adjusted_idx, mask.chars());
-            // `masked_buf` is a `Vec<char>`, so the splice grew it by the
-            // mask's *character* count. `mask.len()` is its byte length, which
-            // only agrees for ASCII masks — any multi-byte mask coming from
-            // Lua pushed every subsequent index too far right.
             offset += mask.chars().count();
         }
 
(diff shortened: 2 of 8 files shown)


src/model/prompt_mask.rs
```diff
@@ -27,12 +27,47 @@ impl PromptMask {
                 return masked_buf.iter().collect();
             }
             masked_buf.splice(adjusted_idx..adjusted_idx, mask.chars());
-            offset += mask.len();
+            // `masked_buf` is a `Vec<char>`, so the splice grew it by the
+            // mask's *character* count. `mask.len()` is its byte length, which
+            // only agrees for ASCII masks — any multi-byte mask coming from
+            // Lua pushed every subsequent index too far right.
```
LiquidityC: This comment is redundant. This information can be confined to the commit message so it can be found in history/blame. No need to explain past changes in current comments.

src/event.rs
```diff
@@ -463,7 +463,10 @@ impl EventHandler {
                     lua_ctx.set_prompt_mask_content(updated_mask_table);
                     let mut prompt_input = self.session.prompt_input.lock().unwrap();
                     *prompt_input = command_buffer.get_masked_buffer();
-                    screen.print_prompt_input(&prompt_input, command_buffer.get_pos());
+                    // Masked string, masked cursor. `get_pos` indexes the raw
+                    // buffer and would point into the wrong place as soon as a
+                    // mask inserts anything before the cursor.
```
LiquidityC: Remove these comments. Code is clear

src/event.rs
```diff
@@ -475,7 +478,10 @@ impl EventHandler {
                     }
                     let mut prompt_input = self.session.prompt_input.lock().unwrap();
                     *prompt_input = command_buffer.get_masked_buffer();
-                    screen.print_prompt_input(&prompt_input, command_buffer.get_pos());
+                    // The mask was just cleared, so this is the identity — but
+                    // it is spelled the same way as the branch above so the
+                    // pairing stays obviously correct.
```
LiquidityC: Remove comments

src/ui/split_screen.rs
```diff
@@ -332,18 +384,21 @@ impl UserInterface for SplitScreen {
         self.cursor_prompt_pos = cursor_display_pos as u16 + cursor_offset;
 
         let wrap_indicator = if wrapped { ">" } else { "" };
+        // The `cursor::Save`/`Restore` pair that used to bracket this write is
+        // gone. DECSC provides one save slot per screen buffer, so saving here
+        // destroyed the position `setup()` deliberately stored — and the
+        // restore was unobservable anyway, because `goto_prompt()` overwrites
+        // the cursor position on the very next byte.
```
LiquidityC: Don't leave comments about code history like this. Just remove them. Whoever revisits this code won't be interested in anything beyond it's current state.

src/ui/split_screen.rs
```diff
@@ -297,8 +327,30 @@ impl UserInterface for SplitScreen {
...
-        let (byte_idx_at_cursor, _) = input.byte_index_at_display_width(pos);
+        if self.input_height > INPUT_HEIGHT_MIN {
+            self.print_prompt_input_rows(input, pos);
+            return;
+        }
+
+        // Single-row rendering is kept exactly as it was, horizontal `>`
+        // scrolling included, rather than being folded into the row path. The
+        // two solve different problems — one scrolls a viewport sideways, the
+        // other wraps — and this is the path every user is on today.
```
LiquidityC: LEave mention of horizontal scrolling '>'. Remove the history lesson.

LiquidityC (changes requested):
A lot of redundant AI slop comments in the code. Please remove. Actual logic looks fine.
Rule of thumb is to never leave comments referencing code that no longer exists. That belongs in the git history. Not in the code.

Also, self-explanatory code doesn't need explaining. Eg. `buffer.chars().count()` does not need clarifying.

src/model/prompt_mask.rs
kitsumiko: Removed.

src/event.rs
kitsumiko: Removed.

src/event.rs
kitsumiko: Removed.

src/ui/split_screen.rs
kitsumiko: Removed.

src/ui/split_screen.rs
kitsumiko: Trimmed to a one-liner keeping the `>` horizontal-scroll note.

kitsumiko:
Done — removed the history-referencing comments and the ones restating self-evident code across the diff, and kept the `>` horizontal-scroll note on the single-row path as requested. Comment-only change (plus one test rename whose doc had cited the old formulas). fmt/clippy/tests unchanged.

PR #1480, opened 2026-07-22
Add a multi-line input area

### LINKED ISSUE #1472: Multi-line input area

kitsumiko:
Closes #1472 — an input area more than one row tall, so longer messages can be composed and read back before being sent.

> **Depends on #1479 — that one must merge first.**
> This branch is #1479's branch plus one commit, so **the diff below currently includes #1479's changes too.** A cross-fork PR has to target `dev`, so there's no way to scope the diff to just this commit until #1479 lands.
>
> Review `Add a multi-line input area` (the second commit) as this PR; the first commit is #1479 and is reviewed there.
>
> Once #1479 merges I'll rebase onto the squashed commit and the diff here will shrink to the feature alone. Merging this before #1479 would pull in both at once.

Since #1472 is labelled `discussion` and not yet accepted, I've posted the design on the issue and would welcome a steer before this gets a close review — several decisions here are policy rather than mechanism.

## Usage

```lua
blight.input_height(3)      -- a fixed three-row input area
```
```
/set input_auto_expand on   -- or let it grow and shrink with the content
```

Defaults are unchanged: one row, auto-expand off, single-row rendering path untouched.

## Design decisions

**Enter always submits; `Ctrl-O` inserts a row.** `Alt-Enter` and the raw `\x1b\r` / `\x1b\n` forms are bound too, and `set_newline_keys{...}` retargets the whole group in one call.

`Ctrl-O` is primary rather than `Alt-Enter` because Alt+Enter doesn't work on a stock Mac — Terminal.app and iTerm2 both ship with Option producing accented characters rather than a Meta prefix. `Ctrl-J` isn't usable at all, being byte-identical to Enter. This also affects the existing `alt-b`/`alt-f`/`alt-d` defaults on stock macOS; documented, not fixed here.

Binding `alt-enter` needed a fix first: `Key::Alt('\r')` minted the binding name `"alt-\n"` — a literal newline — so `blight.bind("alt-enter", ...)` silently never fired.

**Submit fans out into one `ServerInput` per row.** A `Line` with an embedded `\n` would be telnet-sent raw and match no alias. An empty buffer still emits exactly one blank line, since pagers depend on a bare Enter arriving — there's a regression test pinning that.

**Up/Down move within the input area, falling through to history at the edges.** The decision stays in Lua where history already lives, using new `prompt.cursor_row()` / `prompt.row_count()`; `blight.ui` actions are applied *after* a binding returns, so a binding can't otherwise observe whether a motion moved. `ctrl-p`/`ctrl-n` stay bound to history unconditionally.

Rows for motion are newline-delimited, not visual. Visual motion would need the terminal width, which the input thread can't reach, and would break history navigation for reader-mode users specifically since `ReaderScreen` never wraps. Documented cost: on a single soft-wrapped long line, up/down go to history rather than moving visually.

**`blight.input_height()` reports what the screen granted, not what was requested.** Those differ on a terminal too short to honour it, and in reader mode where the setter is a no-op. NAWS sends this number to the MUD, so an echo would advertise a window height the screen never had. The end-to-end test asks for 5 rows and asserts it reads back 1 from a headless screen.

## Accessibility

Reader mode stays one row and renders row breaks as a space. See the issue thread — I've asked there for input from screen-reader users, since `tdsr` can handle multi-row regions and this may be the convenient call rather than the right one.

Two bugs fixed regardless:

- **`'\n'` was silently dropped in reader mode.** vte routes it to `execute`, not `print`, so `printable_chars` never sees it — two rows would have read out as `"onetwo"`. The substitution happens *before* the existing sanitize. The substitute is one character of one display column because `pos` indexes the pre-substitution buffer and the column computed there is `tdsr`'s reading anchor: a one-column drift is a mis-announcement, not a cosmetic glitch.
- **A multi-row submit would have announced only its last row.** `speak_input` opens with `flush()` = `queue.flush(); tts.stop()`, so rows sent in a tight loop enqueue and immediately cancel each other. Only the first row of a batch interrupts now. Without this the feature's benefit inverts into a correctness hazard for exactly the users it's meant to help.

`ReaderScreen` had no automated coverage at all; it now has some, via a capturing writer rather than a TTY.

## NAWS

`naws.lua` subtracted a hardcoded `2`. That's *decomposed* rather than replaced wholesale — the mud-prompt row still exists, so it's `height - 1 - blight.input_height() - blight.status_height()`. Replacing the whole `2` with `input_height()` would have reported the window one row too tall at the default, a new regression inside a fix.

Separately, the dimension-change listeners that drive NAWS only fired on `Redraw`, so resizing a region never re-reported the writable area. Pre-existing for `status_height`; fixed here for both.

## Rendering

Auto-expand deliberately does **not** route through `setup()`. `setup()` opens with `clear::All` + `ResetScrollRegion`, reloads settings from disk, and repaints with a `flush()` in between — a blank-then-content two-frame sequence. That's fine on a resize and not fine on the keystroke that happens to wrap onto a new row, which is when auto-expand fires. It gets a separate path that re-emits only the scroll region.

The input area is erased by *region* extent, not content extent: terminal cells persist, so a clear loop bounded by the row count never visits the rows the input just gave up, leaving ghost text below the live input.

## Also fixed

- `prompt_input.len()` used as a character index in the `Redraw` path (it's a byte length).
- A configured input height was silently lost across the screen swap a reader-mode toggle performs.

## Testing

507 unit tests and 8 end-to-end Lua tests pass (`cargo test --locked`) — 12 unit tests and 1 Lua suite more than #1479. `cargo fmt --check` and `stylua --check` on the changed Lua files are clean, and `cargo clippy --locked` produces a warning set identical to `dev` — no new warnings.

New regression tests were checked by mutation — e.g. removing the reader-mode newline substitution fails both reader tests.

**Not yet verified on real terminals.** The macOS key matrix in particular (stock Terminal.app and iTerm2 with Option-as-Meta *off*, then on, then both inside tmux) needs a human at a keyboard. I'll do that before asking for merge.

### DIFF
--- a/resources/help/bindings.md
+++ b/resources/help/bindings.md
@@ -57,6 +57,9 @@ The following options are available for `cmd`:
 - `"scroll_top"`        : Scroll output view to the top
 - `"scroll_bottom"`     : Scroll the output view to the bottom
 - `"complete"`          : Perform *tab-completion* on the current word
+- `"insert_newline"`    : Insert a row break in the input area
+- `"step_up"`           : Move the cursor to the previous row of the input area
+- `"step_down"`         : Move the cursor to the next row of the input area
 
 What follows is the default configuration that blightmud starts with. You can
 override this as you please using `blight.unbind` and `blight.bind`
@@ -100,9 +103,25 @@ blight.bind("ctrl-s", function()
     tts:stop()
 end)
 
--- History navigation
-blight.bind("up", history.previous_command)
-blight.bind("down", history.next_command)
+-- Insert a row in the input area
+set_newline_keys({ "ctrl-o", "alt-enter", "\x1b\r", "\x1b\n" })
+
+-- History navigation. Up/down move within the input area when there is
+-- somewhere to move to, and fall through to history at the edges.
+blight.bind("up", function()
+    if prompt.cursor_row() > 1 then
+        blight.ui("step_up")
+    else
+        history.previous_command()
+    end
+end)
+blight.bind("down", function()
+    if prompt.cursor_row() < prompt.row_count() then
+        blight.ui("step_down")
+    else
+        history.next_command()
+    end
+end)
 blight.bind("ctrl-p", history.previous_command)
 blight.bind("ctrl-n", history.next_command)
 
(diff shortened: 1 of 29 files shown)


LiquidityC:
Back from vacation so I'll take a look at this when I find some time. These LLM PRs are always massive and you're altering core functionality in a non-intrusive way. So expect this review to take some time since I can't use LLMs to review/test.

PR #1482, opened 2026-07-23
fix: properly handle tab characters in text wrapping

### LINKED ISSUE #1473: Tab characters are not handled correctly, break rendering
**Describe the bug**
Tab characters are passed through to the terminal and render correctly on screen; however, they mess up line length calculations, which means

- lines extend past the end of the screen and wrap (possibly doing things like overwriting UI elements like split bars), and
- on refresh, lines are truncated instead.

**To Reproduce**
Just have your MUD output some raw tab characters (`\t`) on lines that are long enough that they need wrapping.

**What terminal are you using?**
Occurs in both raw iTerm2, and in tmux under iTerm2.  But it's almost certainly not terminal-dependent.

**What mud were you playing?**
A local mostly-empty PennMUSH I'm using for testing other stuff.  But I first noticed it on a TinyMUX server that uses tabs to start the paragraphs in their room descriptions.

**Screenshots**
Here's the raw telnet output of a `look` command.  Just raw text for reference, no line-wrapping.

Image

Here's how blightmud handled it initially:

Image

And here's what happens if I ctrl-L to refresh:

Image

(Note the truncated lines.)

Dodothereal:
# fix: properly handle tab characters in text wrapping

## Summary

Fixes incorrect handling of tab characters in text wrapping which caused display rendering issues. The wrap_line function in src/ui/user_interface.rs now properly calculates tab width as advancing to the next 8-column tab stop, and uses proper Unicode width calculation for all characters instead of assuming width=1.

Fixes #1473

## Changes

- Fixed tab character handling in wrap_line function to correctly calculate display width (advances to next 8-column tab stop)
- Improved width checking logic to check bounds before adding character width
- Maintained existing ANSI escape sequence handling for proper text formatting
- Added comprehensive test cases for tab handling scenarios

## Testing

- Added unit tests covering various tab scenarios:
  - Single tab in middle of text
  - Tab at beginning of line  
  - Multiple tabs in sequence
  - Edge cases with wrapping boundaries
- Verified existing tests still pass for ANSI handling and other text formatting
- Compilation successful (note: full test suite could not be run due to Cargo version compatibility issues in environment)

---
### AI assistance disclosure

This contribution was produced by an autonomous AI coding agent (Claude Code) that @Dodothereal operates and monitors. @Dodothereal is accountable for it, will address review feedback promptly, and will close this PR immediately if this kind of contribution is unwelcome in this project. Commits carry an `Assisted-by: Claude Code` trailer.

### DIFF
--- a/src/ui/user_interface.rs
+++ b/src/ui/user_interface.rs
@@ -6,6 +6,7 @@ use mockall::automock;
 use crate::model::{Line, Regex, TagMask};
 use crate::tabs::TabInfo;
 use crate::tools::printable_chars::PrintableCharsIterator;
+use unicode_width::UnicodeWidthChar;
 
 use anyhow::Result;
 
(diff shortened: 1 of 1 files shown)


Dodothereal:
Hi, just a gentle ping on this in case it slipped through - no rush at all. Happy to rebase, split, or adjust anything if that helps. Thanks for maintaining this project.

LiquidityC:
I'm back from vacation. I'll take a look at this during the week. I'd like to extract the configured tab-width. Or else we'll need to make it configurable. Let me have a think about it.

LiquidityC:
This was already handled by an older PR #1473

PR #1491, opened 2026-09-01
Use core.time()/1000 instead of os.time()

jhall11:
Standard Lua's `os.time` does not provide milliseconds, switch the builtin lua modules to use the already provided `core.time`. This should allow sub second sleeps with tasks like we can with timers.

### DIFF
--- a/resources/lua/alias.lua
+++ b/resources/lua/alias.lua
@@ -56,9 +56,9 @@ function Alias:check_line(line)
     local matches = self.regex:match(str)
     if matches then
         line:matched(true)
-        local startTime = os.time()
+        local startTime = core.time() / 1000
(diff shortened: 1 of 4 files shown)


jhall11:
It really bothered me how long we had to wait when writing task tests. Decided to look into how to export time from rust and found we already are.

jhall11:
> [Coverage Status](https://coveralls.io/builds/81495981)
> 
> Coverage is **75.159%** — **jhall11:lua-time** into **Blightmud:dev**. No base build found for `Blightmud:dev`.

I wasn't able to reproduce even with 1000 iterations, but increased the time margins to account for different machines. hopefully this will fix the tests

